PR #3, opened 2018-06-25
Change the version comparision syntax, foo|version_compare is deprecated

sac:
Replace version_compare with version(), version_compare was renamed in 2.5.
ref: https://docs.ansible.com/ansible/2.5/user_guide/playbooks_tests.html#version-comparison


tyll:
Thank you for your contribution!

PR #10, opened 2020-12-01
Fake pr for debugging https://github.com/linux-system-roles/test-harness/issues/124

nhosoi:
CI tests with collections option fail with "__spec__ is None" if the role has module_utils #124


nhosoi:
centos-8/ansible-2.9 (staging) — The role does not support this platform. Skipping.

PR #14, opened 2021-01-11
firewall zone support with firewalld

t-woerner:
The firewall role has been extended to be able to support the definition
of the zone a change is ensured on. A given zone will be used for these
parameters: `service`, `port` and `forward_port` without a given interface
or MAC address.

Also minor code cleanup has been done to enable the use of zones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,14 @@ firewall_setup_default_solution: false
 
 This turns off the installation and start of the default firewall solution for the specific Fedora or RHEL release. This is intended for users of system-config-firewall on RHEL-7 or Fedora releases.
 
+### zone
+
+Name of the zone that should be modified. The zone parameter is only supported with firewalld. If it is not set for firewalld, the default zone will be used. It will have an effect on these parameters: `service`, `port` and `forward_port` without a given interface or MAC address.
+
+```
+zone: 'public'
+```
+
 ### service
 
 Name of a service or service list to add or remove inbound access to. The service needs to be defined in firewalld or system-config-firewall/lokkit configuration.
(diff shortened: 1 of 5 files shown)


library/firewall_lib.py
```diff
@@ -82,6 +82,12 @@
         <mac-addr>;<port>[-<port>]/<protocol>;[<to-port>];[<to-addr>].
     required: false
     default: null
+  zone:
+    description:
+      Name of the zone that the change will be done one for firewalld.
+      If the zone name is not given, then the default zone ull be used.
```
nhosoi, line 88: ^ull^null^ ?

t-woerner: FIxed

README.md
```diff
@@ -206,6 +218,24 @@ It is also possible to combine several settings into blocks:
...
+  vars:
+    firewall:
+      - { zone: 'external',
+          service: [ 'tftp', 'ftp' ],
+          port: [ '443/tcp', '443/udp' ],
+          forward_port: [ '447/tcp;;1.2.3.4',
+                          '448/tcp;;1.2.3.5' ],
+          state: 'enabled' }
+  roles:
+    - firewall
```
richm, line 234: Should use the FQRN - `linux-system-roles.firewall`

tests/tests_zone.yml
```diff
@@ -0,0 +1,125 @@
...
+        shell:
+          cmd: |
+            firewall-cmd --permanent --load-zone-defaults=internal
+            firewall-cmd --permanent --load-zone-defaults=external
+            firewall-cmd --permanent --load-zone-defaults=trusted
+            firewall-cmd --reload
+
+    when: ansible_distribution == "Fedora" or
+      ((ansible_distribution == "CentOS" or ansible_distribution == "RedHat")
+       and ansible_distribution_major_version >= "7")
```
richm, line 125: `and ansible_distribution_major_version|int >= 7` - to use numeric comparison instead of string (but string is ok for this particular case)

t-woerner: FIxed

README.md
t-woerner: Fixed

PR #32, opened 2021-08-02
Remove support for lokkit, drop NetworkManager interaction

t-woerner:
- Remove support for system-config-firewall/lokkit
- Remove firewall connection support
- Remoce MAC address support
- Remove interface, NetworkManager, trust, masq support
- Cleanup unused imports, move AnsibleModule import after docs


library/firewall_lib.py
```diff
@@ -19,10 +19,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from __future__ import absolute_import, division, print_function
-
-__metaclass__ = type
-
```
richm, line 25: We need to keep this boilerplate

PR #33, opened 2021-08-04
Add support for ansible.posix.firewalld options

t-woerner:
Support for masquerade, rich_rule, source, interface, icmp_block,
icmp_block_inversion, timeout, target, permanent, runtime (alias:
immediate) and offline have been added.

The forward_port parameter is using the string format that firewalld is
using.

Additionally the source_port parameter has been added.

The new test tests_ansible.yml has been added. It is importing the role and
using the firewall_lib module directly.

This PR bases on #32


richm:
> Hi! I'm currently looking this over, the build is failing- would you mind fixing it?

If you are talking about the `tox / python (3.8)` test, then you can ignore that - it's failing during the molecule test - it's going to take some work to make our tests work with molecule

So, lgtm

PR #37, opened 2021-09-10
parametrize module parameter testing; enable coverage

richm:


richm:
@vrindle

vrindle:
Looks good. It seems like we can merge.

PR #38, opened 2021-09-13
Accept install git in workflows

patsevanton:


patsevanton:
If use `sudo apt-get install git` will be error

richm:
superseded by https://github.com/linux-system-roles/firewall/pull/39

PR #40, opened 2021-09-15
[Draft] added changes so present and absent can be added to zones.

vrindle:


library/firewall_lib.py
```diff
@@ -299,6 +301,29 @@ def main():
...
+            + len(source_port)
+            + len(forward_port)
+            + len(rich_rule)
+            + len(source)
+            + len(interface)
+            + len(icmp_block)
+            > 0
+        ):
+            module.fail_json(
+                msg="The states present and absent can only be used in zone level operations (i.e. when no other parameters but zone and state are set)."
```
t-woerner, line 324: In my opinion `present` or `enabled` should always ensure that a zone exists. If the zone does not exist, it should simply create a zone with the given settings.
"absent" with zone parameters should ensure the absence of the given settings within the zone. Without zone parameters it should ensure the absence of the zone. To have a better differentiation about the presence and absence of items in lists it might be good to have an additional parameter. In ansible-freeipa for example we are using the `action` parameter for this. (Example: https://github.com/freeipa/ansible-freeipa/blob/master/README-hostgroup.md)

library/firewall_lib.py
```diff
@@ -299,6 +301,29 @@ def main():
             "icmp_block_inversion, target or zone needs to be set"
         )
 
+    if state == "present" or state == "absent":
```
t-woerner, line 304: What about using `present` to make sure that a zone exists exactly with the given settings? `enabled` and `disabled` can be used as is to make sure that an items exist or do not exist in the settings.

richm:
@t-woerner this is for feature parity with ansible.posix.firewalld - https://docs.ansible.com/ansible/latest/collections/ansible/posix/firewalld_module.html#parameter-state
"The states present and absent can only be used in zone level operations (i.e. when no other parameters but zone and state are set)."

richm:
I think this PR can be closed

PR #47, opened 2021-10-06
Made interface require zone, target require zone, and source require …

vrindle:
…permanent.

Target cannot only be set when set when state is present or absent.


richm (approved):
lgtm

PR #48, opened 2021-10-07
Update interface so that interface can only be applied to one zone.

vrindle:


richm:
[citest pending]

richm:
[citest pending]

library/firewall_lib.py
```diff
@@ -178,7 +178,39 @@
 except ImportError:
     HAS_FIREWALLD = False
 
-
+def handle_interface(zone,item,fw_zone,fw_settings,fw,fw_offline):
```
richm, line 181: use `interface` instead of `item`

library/firewall_lib.py
```diff
@@ -178,7 +178,39 @@
...
+    if fw_offline:
+        iface_zone_objs = []
+        for zone in fw.config.get_zones():
+            old_zone_obj = fw.config.get_zone(zone)
+            if item in old_zone_obj.interfaces:
+                iface_zone_objs.append(old_zone_ob)
```
stepnem, line 187: old_zone_obj?

(Not really reviewing this, Rich's comments just landed in my inbox and this caught my eye.)

Also, I'm surprised this seems to have passed the checks.
Don't you run some kind of static analysis? pylint?

library/firewall_lib.py
```diff
@@ -178,7 +178,39 @@
...
+            fw_settings.addInterface(item)
+            fw.config.set_zone_config(fw_zone, fw_settings.settings)
+    else:
+        old_zone_name = fw.config().getZoneOfInterface(item)
+        if old_zone_name != zone:
+            if old_zone_name:
+                old_zone_obj = fw.config().getZoneByName(old_zone_name)
+                old_zone_settings = old_zone_obj.getSettings()
+                old_zone_settings.removeInterface(item)
+                fw_zone.update(fw_settings)
```
erig0, line 212: This `update()` looks redundant. This file calls `update()` once at the end already.

library/firewall_lib.py
```diff
@@ -178,7 +178,39 @@
...
+        for zone in fw.config.get_zones():
+            old_zone_obj = fw.config.get_zone(zone)
+            if item in old_zone_obj.interfaces:
+                iface_zone_objs.append(old_zone_ob)
+        if len(iface_zone_objs) > 1:
+            module.fail_json(
+                msg="ERROR: interface {0} is in {1} zone XML file, can only be in one".format(
+                    interface, len(iface_zone_objs)
+                )
+            )
```
erig0, line 193: This should never happen. If we want to enforce the "an interface is bound to one and only one zone" rule in this role then why can't we just remove the interface from all zones other than the new zone? No need to throw an error and give up. This will also simplify the surrounding code.

tests/unit/test_firewall_lib.py
```diff
@@ -179,7 +179,7 @@
         "enabled": {
             "expected": {
                 "runtime": [call("default", "eth2")],
-                "permanent": [call("eth2")],
+                "permanent": [],
```
erig0, line 182: I can't tell if this change is intentional or not. Pinging it just in case.

library/firewall_lib.py
```diff
@@ -683,7 +715,7 @@ def exception_handler(exception_message):
                 changed = True
```
richm, line 715: For runtime, this code uses `fw.addInterface(zone, interface)`
however, the a.p.firewalld code uses `fw.changeZoneOfInterface(zone, interface)` instead - https://github.com/ansible-collections/ansible.posix/blob/main/plugins/modules/firewalld.py#L459

Which one is correct? @erig0 or @t-woerner ?

library/firewall_lib.py
```diff
@@ -179,6 +179,40 @@
...
+):
+    if fw_offline:
+        iface_zone_objs = []
+        for zone in fw.config.get_zones():
+            old_zone_obj = fw.config.get_zone(zone)
+            if item in old_zone_obj.interfaces:
+                old_zone_settings = FirewallClientZoneSettings(
+                    fw.config.get_zone_config(old_zone_obj)
+                )
+                old_zone_settings.removeInterface(item)
```
richm, line 193: You don't need line 201 anymore since you are doing the `removeInterface` on line 193 and that should catch all zones.
I think line 202 should be moved here:
```
             fw.config.set_zone_config(old_zone_obj, old_zone_settings.settings)
```

vrindle: If having both `removeInterface` is all that is needed, are we all set to merge this change in?

PR #58, opened 2021-12-06
Added examples of options in Readme

vrindle:


README.md
```diff
@@ -224,6 +224,82 @@ permanent: true
 The permanent and runtime settings are independent, so you can set only the runtime, or only the permanent.  You cannot
 set both permanent and runtime to `false`.
 
+Examples of Options
+-------------------
```
richm, line 228: Since the default is `permanent: yes` and `runtime: yes` I think you can omit them from the examples, and say something like "By default, any changes will be applied immediately, and to the permanent settings.  If you want the changes to apply immediately but not permanently, use `permanent: no`.  Conversely, use `runtime: no`".

README.md
```diff
@@ -224,6 +224,82 @@ permanent: true
...
+
+Permit TCP traffic for port 80 in permanent default zone:
+
+```yaml
+firewall:
```
richm, line 233: The firewall role `firewall` variable value is a list - so this should be
```yaml
firewall:
  - port: 80/tcp
    state: enabled
```

I note that there are some corrections we need to make in the role README as well - like https://github.com/linux-system-roles/firewall#state
https://github.com/linux-system-roles/firewall#example-playbooks
and in "It is also possible to combine several settings into blocks:" there are a couple of extraneous dashes `-` in the left hand column.

Also, the README https://github.com/linux-system-roles/firewall#variables section should start off with something like 
```
The firewall role uses the variable `firewall` to specify the parameters.  This variable is a `list` of `dict` values.  Each `dict` value is comprised of one or more keys listed below.
```

README.md
```diff
@@ -224,6 +224,77 @@ permanent: true
...
+  - port: 80/tcp
+    state: disabled
+```
+
+Add masquerading to dmz zone:
+
+```yaml
+firewall:  
+  - masquerade: yes
+    zone: dmz
```
richm, line 252: `state` is required - please add `state: enabled`

README.md
```diff
@@ -224,6 +224,77 @@ permanent: true
...
+```
+
+Remove masquerading to dmz zone:
+
+```yaml
+firewall:
+  - masquerade: no
+    zone: dmz
```
richm, line 260: `state` is required - please add `state: disabled`

README.md
richm, line 252: `state` is a required parameter - please add `state: enabled` here

README.md
richm, line 260: `state` is a required parameter - please add `state: disabled` here

vrindle: I think the state would have to be set to `state: enabled here` since masquerading cannot be used with the `state` being set to `disabled`. https://github.com/linux-system-roles/firewall/blob/94e0c6dde38200f3db33ae069388a3075ee1a7d9/library/firewall_lib.py#L402

richm (approved):
lgtm

PR #59, opened 2021-12-14
Added implicit firewalld reload for when a custom zone is added or re…

vrindle:
…moved

### DIFF
--- a/tests/tests_ansible.yml
+++ b/tests/tests_ansible.yml
@@ -333,12 +333,33 @@
           zone: customzone
           state: present
           permanent: yes
+          immediate: yes
         register: result
 
       - name: assert firewalld custom zone
         assert:
           that:
             - result is changed
+
+      - name: Add masquerade to custom zone
+        firewall_lib:
+          masquerade: yes
+          zone: customzone
+          state: enabled
+          permanent: yes
+          immediate: yes
+        register: result
+        failed_when: result is failed or not result is changed
+
+      - name: Add masquerade to custom zone, again
+        firewall_lib:
+          masquerade: yes
+          zone: customzone
+          state: enabled
+          permanent: yes
+          immediate: yes
+        register: result
+        failed_when: result is failed or result is changed
 
     always:
 
--- a/tests/tests_zone.yml
+++ b/tests/tests_zone.yml
@@ -85,7 +85,7 @@
             - zone: customzone
               permanent: yes
               masquerade: yes
-              state: enabled 
+              state: enabled
 
       - name: Fail on newly changes
         fail: msg="FAILED"
@@ -114,7 +114,7 @@
                      or "port=447:proto=tcp:toport=:toaddr=1.2.3.4"
                         not in result.stdout
                      or "port=448:proto=tcp:toport=:toaddr=1.2.3.5"
-                          not in result.stdout
+                        not in result.stdout
 
       - name: Verify custom zone has masquerade added to it
         command: firewall-cmd --permanent --zone=customzone --query-masquerade
--- a/tests/tests_zone.yml
+++ b/tests/tests_zone.yml
@@ -41,6 +41,19 @@
               forward_port: [ '447/tcp;;1.2.3.4',
                               '448/tcp;;1.2.3.5' ]
               state: 'enabled'
+            - zone: 'internal'
+              service: [ 'tftp', 'ftp' ]
+              port: [ '443/tcp', '443/udp' ]
+              forward_port: [ '447/tcp;;1.2.3.4',
+                              '448/tcp;;1.2.3.5' ]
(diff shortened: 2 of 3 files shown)


tests/tests_ansible.yml
```diff
@@ -333,13 +333,34 @@
...
 
+      - name: Add masquerade to custom zone
+        firewall_lib:
+          masquerade: yes
+          zone: customzone
+          state: enabled
+          permanent: yes
+          immediate: yes
+        register: result
+        failed_when: result.failed or not result.changed
```
richm, line 352: should use an ansible test for this: https://docs.ansible.com/ansible/2.9/user_guide/playbooks_tests.html#task-results
`failed_when: result is failed or not result is changed`

richm:
I think there should be a test of the role itself which shows adding a zone then adding something to the zone in a single firewall role invocation, then verify that the zone and the setting is present, then remove the setting and the zone in a single firewall role invocation, and verify the zone is removed.  Maybe you can change some existing tests in tests/tests_zone.yml, or add a new test.  Not sure how zones work - maybe you can remove a zone that has settings in it?  For example, if I create zone `foo` and enable service https, then I remove zone `foo`, I'm assuming I don't need to first disable https.

There should also be some examples showing how to add a zone and how to add settings for the zone in the same firewall role invocation.

tests/tests_ansible.yml
richm: hmm - looking at some of the existing tests - they all use the `result.failed` style instead of `result is failed` - I think you should not go back and change all of them, but if that code is changed in the future for some reason, it should also change the tests to use the `result is failed` style.

richm:
you can ignore the CentOS-7-latest/ansible-2.9 and RHEL-7.9-20200917.0/ansible-2.9 errors - so the tests look good

tests/tests_ansible.yml
```diff
@@ -333,13 +333,34 @@
...
 
+      - name: Add masquerade to custom zone
+        firewall_lib:
+          masquerade: yes
+          zone: customzone
+          state: enabled
+          permanent: yes
+          immediate: yes
+        register: result
+        failed_when: result is failed or not result.changed
```
richm, line 352: `result is failed or not result is changed` https://docs.ansible.com/ansible/2.9/user_guide/playbooks_tests.html#task-results

tests/tests_ansible.yml
```diff
@@ -333,13 +333,34 @@
...
+
+      - name: Add masquerade to custom zone, again
+        firewall_lib:
+          masquerade: yes
+          zone: customzone
+          state: enabled
+          permanent: yes
+          immediate: yes
+        register: result
+        failed_when: result.failed or result.changed
```
richm, line 362: `result is failed or result is changed`

tests/tests_zone.yml
```diff
@@ -41,6 +41,10 @@
                 forward_port: [ '447/tcp;;1.2.3.4',
                                 '448/tcp;;1.2.3.5' ],
                 state: 'enabled' }
+            - { zone: 'customzone',
```
richm, line 44: unless the value is a single line for compactness, it is better to use the standard yaml layout for a hash value instead of the json style - so
```
  - zone: customzone
    masquerade: yes
    permanent: yes
    state: enabled
```
I see that the other values in tests_zone use `{` and `}` but they shouldn't 
NOTE that in some cases using the json style is a good thing e.g. for the array values like
```
                service: [ 'tftp', 'ftp' ],
                port: [ '443/tcp', '443/udp' ],
```
instead of the yaml style
```
                service:
                  - tftp
                  - ftp
                port:
                 - '443/tcp'
                 - '443/udp'
 ```
because the json style makes the code more readable

also note that there is a big difference between `param: yes` and `param: 'yes'` - the former is a boolean value, and the latter is a string value - they are not the same value.  Since parameters like `masquerade` and `permanent` are boolean valued, you must use `yes` and not `'yes'`

File: tests/tests_zone.yml [hunk 5]
```diff
@@ -88,7 +96,13 @@
                      or "port=447:proto=tcp:toport=:toaddr=1.2.3.4"
                         not in result.stdout
                      or "port=448:proto=tcp:toport=:toaddr=1.2.3.5"
-                        not in result.stdout
+                          not in result.stdout
+
+      - name: Verify custom zone has masquerade added to it
+        command: firewall-cmd --permanent --zone=customzone --queryMasquerade
+        register: result
+        failed_when: result.failed
```
richm, line 104: `result is failed or ...`

richm:
@vrindle lgtm - do you want to remove the draft status?  is it ready for review?

richm (approved):
lgtm

PR #68, opened 2022-02-15
Add ability to set the default zone

vrindle:
WIP


library/firewall_lib.py
```diff
@@ -251,6 +256,9 @@ def parse_forward_port(module, item):
     return (_port, _protocol, _to_port, _to_addr)
 
 
+def set_the_default_zone(fw, set_default_zone):
```
richm, line 259: I don't think we need a function for this - just use `fw.setDefaultZone(set_default_zone)` below

library/firewall_lib.py
```diff
@@ -535,6 +546,12 @@ def exception_handler(exception_message):
             fw_zone = None
             fw_settings = None
 
+    # set default zone
+    if set_default_zone:
+        if fw.getDefaultZone is not set_default_zone:
+            set_the_default_zone(fw,set_default_zone)
```
richm, line 552: just use `fw.setDefaultZone(set_default_zone)`

library/firewall_lib.py
```diff
@@ -535,6 +546,12 @@ def exception_handler(exception_message):
...
+        changed = True
```
richm, line 553: This needs to be inside the `if fw.getDefaultZone...` block

tests/tests_ansible.yml
```diff
@@ -398,7 +410,7 @@
       - name: Reset default zone to defaults
         shell:
           cmd: |
-            zone=$(firewall-cmd --get-default-zone)
+            zone=$(public)
```
richm, line 413: do you want this to be the string "public"?  If so, then `zone=public`

tests/tests_zone.yml
```diff
@@ -86,6 +87,7 @@
               permanent: yes
               masquerade: yes
               state: enabled
+            - set_default_zone: dmz
```
richm, line 90: In addition - I suggest adding a test like this:
```
firewall:
  - set_default_zone: somezone
  - service: http
```
the `service` without a zone should be added to the default zone, which should be the one set by the prior item

richm:
[citest]
CI test runners are busy - it make take a couple of hours

richm (approved):

PR #70, opened 2022-02-22
ensure that changes to target take effect immediately

richm:
When changing the target, a reload is required in order for the
changes to take effect, so ensure that this happens.

The code has been refactored slightly in order to use the same
reload logic for other cases where reload is required.

A new test was added for the target use cases.


vrindle:
LGTM

richm:
[citest bad]

PR #72, opened 2022-04-27
Added ability to restore Firewalld defaults

richm:
The user can add `previous: replaced` to one of the elements in
the `firewall` list to force the role to remove all existing
config files.  You can use this to make your Ansible inventory
the "source of truth" for all firewall configuration.

There is a lot of stuff here to ensure that:
* all operations of the role are idempotent
* the role will only report changed: true when something
  actually changed

Specifically - if using `previous: replaced`, the role will only
report `changed: true` if the final configuration, after erasing
and applying all of the configuration, is different than the
initial configuration.


richm:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

richm:
The fedora-35 problem is upstream - their releases are broken - just ignore it

vrindle:
LGTM

PR #74, opened 2022-05-09
remove customzone zone in cleanup

richm:


nhosoi:
Could you add the similar cleanup to `tests_zone.yml`, as well?
```
diff --git a/tests/tests_zone.yml b/tests/tests_zone.yml
index 5873b68..fefee05 100644
--- a/tests/tests_zone.yml
+++ b/tests/tests_zone.yml
@@ -145,6 +145,11 @@
 
       # CLEANUP: RESET TO ZONE DEFAULTS
 
+      - name: Remove customzone zone
+        command: firewall-cmd --permanent --delete-zone=customzone
+        register: result
+        failed_when: result.failed and "INVALID_ZONE" not in result.stderr
+
       - name: Reset to zone defaults
         shell:
           cmd: |
```
And `tests_purge_config.yml` also needs to clean up the `customzone` zone at the end. In addition, in my testing, combined configuration `firewall: "{{ basic_config + [{'previous': 'replaced'}] }}"` does not report `changed`. To make the following verification pass, I had to configure them separately...
```
diff --git a/tests/tests_purge_config.yml b/tests/tests_purge_config.yml
index f694275..7ae7d1b 100644
--- a/tests/tests_purge_config.yml
+++ b/tests/tests_purge_config.yml
@@ -53,11 +53,18 @@
         msg: The role reported changes
       when: firewall_lib_result.changed
 
-    - name: Use previous replaced and basic config
+    - name: Use previous replaced
       include_role:
       vars:
-        firewall: "{{ basic_config + [{'previous': 'replaced'}] }}"
+        firewall:
+          - previous: replaced
+
+    - name: Apply basic config again
+      include_role:
+        name: linux-system-roles.firewall
+      vars:
+        firewall: "{{ basic_config }}"
 
     - name: Verify role reports changes
       fail:
@@ -123,3 +130,10 @@
       fail:
         msg: The role reported changes
       when: firewall_lib_result.changed
+
+    - name: cleanup
+      include_role:
+        name: linux-system-roles.firewall
+      vars:
+        firewall:
+          - previous: replaced
```

richm:
I added proper cleanup blocks to all tests, tagged with `tests::cleanup` so that we can skip if necessary.

The purge_config problem was that I was expecting a change if the role removed the file lockdown-whitelist.xml as part of the test.  However, if `previous: replaced` had been run before running this test, the file would not be there, and the role would report no changes.  In order to avoid this situation, the purge_config test will now wipe out all existing config before running the test to ensure the config is in a known state.

nhosoi:
I ran firewall tests in the serialized manner.
```
:: [ 20:04:16 ] :: [   FAIL   ] :: Test firewall/tests_ansible.yml (roles/rhel-system-roles.firewall/tests/tests_ansible.yml) with ANSIBLE-2 (Expected 0, got 2)
```
It failed here:
```
/usr/share/ansible/roles/rhel-system-roles.firewall/tests/tests_ansible.yml
    364       - name: set default zone
    365         firewall_lib:
    366           set_default_zone: dmz
    367           state: enabled
    368         register: result
    369         failed_when: result is failed or result is not changed
```
```
Ansible output
TASK [set default zone] ********************************************************
task path: /var/tmp/distribution_mcase__root/mcase-relics/mcase/roles/rhel-system-roles.firewall/tests/tests_ansible.yml:364
Monday 09 May 2022  20:04:11 -0400 (0:00:00.357)       0:00:28.776 ************ 
fatal: [myself]: FAILED! => {
    "__firewall_changed": false,
    "changed": false,
    "failed_when_result": true
}
```
What could be wrong in my test env?
```
# firewall-cmd --get-zones
block dmz drop external home internal nm-shared public trusted work
```
Are there any other useful info I could gather? Thanks.

richm:
I guess the default zone was somehow already set to dmz?  Let's see if this fixes it, or if I need to use `firewall-cmd --permanent --set-default-zone public; firewall-cmd --reload`

nhosoi (approved):
Thank you, @richm. This firewall role pr passed the serialized tests.

lgtm

Issue #77, opened 2022-05-27
Use 'runtime' instead of 'immediate' in tests_ansible.yml


richm:
The test gives deprecation warnings because it uses the deprecated `immediate`.  We should change it to use `runtime` instead.

PR #78, opened 2022-05-31
tests_ansible: replaced immediate options with runtime options

BrennanPaciorek:
- Replaced all uses of the depreciated immediate option with the runtime option

Fixes: #77


richm (approved):
lgtm

Issue #76, opened 2022-05-27
State is mandatory for all Firewall Parameters, even for ones that it doesn't make sense for.


vrindle:
**What happened**:
State is currently mandatory even for parameters it doesn't make sense for it. For example icmp_block_inversion and masquerading and setting the default zone should not require the state being enabled but it does.
**What you expected to happen**:
Expectation: For example be able to write the yaml to add masquerading to zone `dmz` WITHOUT having to have state to be set to enabled

**How to reproduce it (as minimally and precisely as possible)**:
N/A
**Anything else we need to know?**:
N/A

PR #79, opened 2022-05-31
fix: state not required for masquerade and ICMP block inversion

BrennanPaciorek:
- The above, and added a new error message for attempting to not specify state when using options that require state like source, port, port_forward

 - Unit test case added for this error message

- Removed state option from integration tests for masquerading and ICMP block, retaining the same fail conditions

Fixes: #76

### DIFF
--- a/library/firewall_lib.py
+++ b/library/firewall_lib.py
@@ -154,7 +154,7 @@ options:
     description:
       Ensure presence or absence of entries.  Use C(present) and C(absent) only
       for zone-only operations, or for target operations.
-    required: true
+    required: false
     type: str
     choices: ["enabled", "disabled", "present", "absent"]
   __report_changed:
@@ -315,7 +315,9 @@ def main():
                 ],
             ),
             state=dict(
-                choices=["enabled", "disabled", "present", "absent"], required=True
+                choices=["enabled", "disabled", "present", "absent"],
+                required=False,
+                default=None,
             ),
             __report_changed=dict(required=False, type="bool", default=True),
         ),
@@ -359,6 +361,19 @@ def main():
     runtime = module.params["runtime"]
     state = module.params["state"]
 
+    # All options that require state to be set
+    state_required = any(
+        (
+            interface,
+            source,
+            service,
+            source_port,
+            port,
+            forward_port,
+            icmp_block,
+            rich_rule,
+        )
+    )
     if permanent is None:
         runtime = True
     elif not any((permanent, runtime)):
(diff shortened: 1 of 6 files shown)


library/firewall_lib.py
```diff
@@ -315,7 +315,9 @@ def main():
                 ],
             ),
             state=dict(
-                choices=["enabled", "disabled", "present", "absent"], required=True
+                choices=["enabled", "disabled", "present", "absent"],
+                required=False,
+                default=None,
```
richm, line 320: I think the default should still be `"enabled"` - this is the general convention in Ansible

richm: hmm - on second thought - adding a default value would be a breaking change to the API - so I think what you have here is correct

richm:
This error: `ERROR: plugins/modules/firewall_lib.py:0:0: doc-required-mismatch: Argument 'state' in argument_spec is not required, but is documented as being required` means you'll need to change the documentation text in the firewall_lib.py code.  

https://github.com/linux-system-roles/firewall/blob/master/library/firewall_lib.py#L153

If you want to run this test locally:

* `dnf install -y podman podman-docker`
* `LSR_ANSIBLE_TEST_DOCKER=true tox -e collection,ansible-test`

richm:
Here is the ansible module dev guide: https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general.html

Specifically: https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general.html#creating-a-module

and https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#developing-modules-documenting

library/firewall_lib.py
BrennanPaciorek: Yeah, I had thought about that, but arrived at the same conclusion.

vrindle:
One of the things that `state` being mandatory for is the `set_default_zone_parameter`. Wherever in the tests `set_default_zone` is enabled, state shouldn't be there. Such as [https://github.com/linux-system-roles/firewall/blob/3b45e1af7b36f63f49c8b2daa00e6bd4171063c1/tests/tests_purge_config.yml#L140](url)
 This should be cleaned up before this is merged.

BrennanPaciorek:
> One of the things that `state` being mandatory for is the `set_default_zone_parameter`. Wherever in the tests `set_default_zone` is enabled, state shouldn't be there. Such as
> 
> https://github.com/linux-system-roles/firewall/blob/3b45e1af7b36f63f49c8b2daa00e6bd4171063c1/tests/tests_purge_config.yml#L142
> 
> This should be cleaned up before this is merged.

This has been fixed

richm:
lgtm - @vrindle ?

vrindle:
LGTM

Issue #80, opened 2022-05-31
Ability to Add User Defined Services


vrindle:
**What would you like to be added**:
Ability to add user defined services using the Firewall System Role.

**Why is this needed**:
It's needed because user's may want to be  able to add their own services rather than the set of services that Firewalld comes with in its defaults.

PR #81, opened 2022-06-03
Add/Delete/Update Services

BrennanPaciorek:
- Users can now add, update, or remove services using states absent and present
   - Absent can only be used to remove ports, modules, protocols, services, destination addresses
   - Present used to add services, short (updates if exists), description (updates if exists), ports, protocols, modules, destination addresses
- Requires permanent option to be true or will show an error instead of setting permanent to true.
   - This is so that people know that what they are doing is permanent and not able to be undone with a reload.
 
 Fixes: #80


BrennanPaciorek:
It looks like all the unit test checks are failing. I'll primarily work on that now. After all the checks are passing, I will move on to adding the last destination address variable.

BrennanPaciorek:
Will remove WIP after CI tests pass

library/firewall_lib.py
```diff
@@ -222,6 +265,43 @@ def parse_port(module, item):
     return (_port, _protocol)
 
 
+ipv4_charset = ".1234567890/"
+ipv6_charset = ":1234567890abcdef/"
+
+
+def parse_destination_address(module, item):
```
richm, line 272: You might also consider using the python ipaddress library https://docs.python.org/3/library/ipaddress.html

library/firewall_lib.py
```diff
@@ -423,10 +525,66 @@ def main():
             )
         ):
             module.fail_json(
-                msg="The states present and absent can only be used in zone level operations (i.e. when no other parameters but zone and state are set)."
+                msg="present and absent only usable for zone or service operations"
```
richm, line 528: I think the more verbose error message is better in this case.  If you are running into issues with flake8 and black and line length, you can wrap like this:
```python
  msg=(
    "The states ......."
    "(i.e. when no ..."
  )
```

library/firewall_lib.py
```diff
@@ -423,10 +525,66 @@ def main():
...
+                    port,
+                    source_port,
+                    helper_module,
+                    protocol,
+                    destination_ipv4,
+                    destination_ipv6,
+                )
+            ):
+                module.fail_json(
+                    msg="short, description, port, source_port, helper_module, "
```
richm, line 551: Does this work?  I thought you had to put this in parentheses like
```python
  msg=(
    "short, description, ..."
    "protocol, or ......"
    "and state ..."
  )
```
?

library/firewall_lib.py
```diff
@@ -423,10 +525,66 @@ def main():
...
+
+    if service_operation:
+        if state == "absent" and any(
+            (
+                description,
+                short,
+            )
+        ):
+            module.fail_json(
+                msg="descrtiption or short is only usable with present state"
```
richm, line 580: description

library/firewall_lib.py
```diff
@@ -579,24 +746,177 @@ def exception_handler(exception_message):
...
 
     # service
-    for item in service:
-        if state == "enabled":
-            if runtime and not fw.queryService(zone, item):
-                if not module.check_mode:
-                    fw.addService(zone, item, timeout)
-                changed = True
-            if permanent and not fw_settings.queryService(item):
+    if service_operation:
```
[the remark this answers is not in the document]
richm: > This service creation and removal code should be refactored into a separate function

library/firewall_lib.py
BrennanPaciorek: It seems to work somwhow

tests/tests_ansible.yml
```diff
@@ -374,14 +374,162 @@
         register: result
         failed_when: result is failed or result is changed
 
+      - name: add multiple custom services
+        firewall_lib:
+          service: custom, another_custom
```
richm, line 379: This will make a service with the name "custom, another_custom" - if you want to pass in a list, you'll need to use a yaml list like
```yaml
  service: [custom, another_custom]
```
or
```yaml
  service:
    - custom
    - another_custom
```
The first method is usually preferred for very short lists like this

library/firewall_lib.py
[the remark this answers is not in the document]
BrennanPaciorek: My main difficulty with this is supporting check mode. I am thinking I can create a mock service object for this. Is there a better way of going about this?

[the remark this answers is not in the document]
richm: > My main difficulty with this is supporting check mode.

What specifically is the problem with check mode?

> I am thinking I can create a mock service object for this. Is there a better way of going about this?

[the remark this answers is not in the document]
BrennanPaciorek: If you are attempting to create a service in check mode, if I understand this correctly, the managed node should not be modified. When you create a service, it directly modifies firewalld's permanent configuration, which is an issue since check mode is not supposed to modify the managed node

My way around this before was to essentially skip the code block once it determines the service specified did not previously exist and set the changed to True, which was not ideal and ended up requiring that I implemented a bunch of redundant code to handle implementing the other service element options in the case that check mode was not on.

Now that I am moving service creation to a separate function, I need to be able to handle service element functions like fw_service_settings.getDescription(), which will break if I do not create the service

TLDR: check mode should not modify managed node, moving service to its own function means that the function needs to modify the system, which contradicts the purpose of check mode.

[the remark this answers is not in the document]
BrennanPaciorek: I can think of a couple of solutions besides a mock object now. one would just be a modification of what I had done before, which is setting changed to True and jumping over the code block if the Ansible is run in check mode and a service would have been created otherwise.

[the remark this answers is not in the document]
richm: > If you are attempting to create a service in check mode, if I understand this correctly, the managed node should not be modified. When you create a service, it directly modifies firewalld's permanent configuration, which is an issue since check mode is not supposed to modify the managed node

Correct.  check mode should not change anything
 
> My way around this before was to essentially skip the code block once it determines the service specified did not previously exist and set the changed to True, which was not ideal and ended up requiring that I implemented a bunch of redundant code to handle implementing the other service element options in the case that check mode was not on.
> 
> Now that I am moving service creation to a separate function, I need to be able to handle service element functions like fw_service_settings.getDescription(), which will break if I do not create the service

I'm still not understanding why creating a separate function for this is more difficult than having the code be in-line in the main function.  Can't you pass in the `fw_service_settings` into the function, or create it in the function and pass it back out to the main function?  Same with the `changed` boolean value and the `check_mode` parameter - pass in the `check_mode` parameter to the function, and have the function return `changed` as one of the return values (a python function can return multiple values i.e. in a tuple).

> 
> TLDR: check mode should not modify managed node, moving service to its own function means that the function needs to modify the system, which contradicts the purpose of check mode.

library/firewall_lib.py
BrennanPaciorek: Do we need to worry about supporting python 2.7? It looks like the ipaddress library was introduced in python 3.3

richm:
Sorry - I forgot about this - ipaddress only takes unicode values - see https://github.com/linux-system-roles/vpn/blob/5e6a33718166e9a89e7d441ce2271cf89a4fa3dd/filter_plugins/vpn_ipaddr.py#L20-L24

richm:
To address the ansible-test import errors - just add an ignore for them to the `.sanity-ansible-ignore-2.*.txt` files - see https://github.com/linux-system-roles/storage/blob/master/.sanity-ansible-ignore-2.12.txt#L1-L8

However, if you feel that using `ipaddress` is more trouble than it's worth, then just use the manual ip address checking

BrennanPaciorek:
> However, if you feel that using `ipaddress` is more trouble than it's worth, then just use the manual ip address checking

Yeah I've decided to use the manual ip checking, since any errors caught beyond an invalid character set will be caught by firewalld's errors

BrennanPaciorek:
> to run local ansible-test tests, use
> 
> ```
> LSR_ANSIBLE_TEST_DOCKER=true tox -e collection,ansible-test
> ```
> 
> This assumes you have done `yum install podman podman-docker`

I do have podman docker, but when running these tests, I get an error that shellcheck is not found. Is this an issue I need to worry about?

richm:
> > to run local ansible-test tests, use
> > ```
> > LSR_ANSIBLE_TEST_DOCKER=true tox -e collection,ansible-test
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This assumes you have done `yum install podman podman-docker`
> 
> I do have podman docker, but when running these tests, I get an error that shellcheck is not found. Is this an issue I need to worry about?
> 
> EDIT: the other tests work just fine, its just that I cannot install shellcheck with dnf or yum.

`yum search shellcheck`?

richm:
looks like shellcheck / ShellCheck package is not available for EL8, neither base nor EPEL8, even though it is in EPEL7 and EL9 :-(

If you want it (you aren't developing any shell code, so you don't really need it), you could go here https://github.com/koalaman/shellcheck#installing and download https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz

richm:
if you want to use `ipaddress` - it only takes unicode values - see https://github.com/linux-system-roles/vpn/blob/5e6a33718166e9a89e7d441ce2271cf89a4fa3dd/filter_plugins/vpn_ipaddr.py#L20-L24 - you'll have to change this:
```python
from __future__ import absolute_import, division, print_function
```
to this
```python
from __future__ import absolute_import, division, print_function, unicode_literals
```

BrennanPaciorek:
> if you want to use `ipaddress` - it only takes unicode values - see https://github.com/linux-system-roles/vpn/blob/5e6a33718166e9a89e7d441ce2271cf89a4fa3dd/filter_plugins/vpn_ipaddr.py#L20-L24 - you'll have to change this:
> 
> ```python
> from __future__ import absolute_import, division, print_function
> ```
> 
> to this
> 
> ```python
> from __future__ import absolute_import, division, print_function, unicode_literals
> ```

It looks like sanity checks dislike when there are any __future__ imports besides absolute_import, division, and print_function are present, and specifically checks that unicode_literals is not imported in sanity checks. I am thinking that if the main Ansible team went out of their way to discourage its use, we're probably better off using the manual checking. Firewalld has its own checks that likely use the ipaddress library that the user can see if they provide bad input.

https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/no-unicode-literals.html

richm:
> > if you want to use `ipaddress` - it only takes unicode values - see https://github.com/linux-system-roles/vpn/blob/5e6a33718166e9a89e7d441ce2271cf89a4fa3dd/filter_plugins/vpn_ipaddr.py#L20-L24 - you'll have to change this:
> > ```python
> > from __future__ import absolute_import, division, print_function
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to this
> > ```python
> > from __future__ import absolute_import, division, print_function, unicode_literals
> > ```
> 
> It looks like sanity checks dislike when there are any **future** imports besides absolute_import, division, and print_function are present, and specifically checks that unicode_literals is not imported in sanity checks. I am thinking that if the main Ansible team went out of their way to discourage its use, we're probably better off using the manual checking. Firewalld has its own checks that likely use the ipaddress library that the user can see if they provide bad input.
> 
> https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/no-unicode-literals.html

Right - for the vpn role, we had no alternative but to use ipaddress, with the unicode-literals, and had to suppress the errors in ansible-test - https://github.com/linux-system-roles/vpn/blob/master/.sanity-ansible-ignore-2.12.txt#L1

However - in your case, just use the manual checking - I think it is fine

BrennanPaciorek:
> > > if you want to use `ipaddress` - it only takes unicode values - see https://github.com/linux-system-roles/vpn/blob/5e6a33718166e9a89e7d441ce2271cf89a4fa3dd/filter_plugins/vpn_ipaddr.py#L20-L24 - you'll have to change this:
> > > ```python
> > > from __future__ import absolute_import, division, print_function
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > to this
> > > ```python
> > > from __future__ import absolute_import, division, print_function, unicode_literals
> > > ```
> > 
> > 
> > It looks like sanity checks dislike when there are any **future** imports besides absolute_import, division, and print_function are present, and specifically checks that unicode_literals is not imported in sanity checks. I am thinking that if the main Ansible team went out of their way to discourage its use, we're probably better off using the manual checking. Firewalld has its own checks that likely use the ipaddress library that the user can see if they provide bad input.
> > https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/no-unicode-literals.html
> 
> Right - for the vpn role, we had no alternative but to use ipaddress, with the unicode-literals, and had to suppress the errors in ansible-test - https://github.com/linux-system-roles/vpn/blob/master/.sanity-ansible-ignore-2.12.txt#L1
> 
> However - in your case, just use the manual checking - I think it is fine

I stuck with the manual checking and just added some more detail to the manual checking as per most recent push to the branch.

richm:
[citest]

richm:
[citest pending]

richm:
[citest pending]

richm:
[citest pending]

richm:
[citest pending]

richm:
[citest bad]

BrennanPaciorek:
The only errors in the CI tests that I can find are `ERROR! Invalid callback for stdout specified: debug`. Is there an issue with my feature that needs resolved?

richm:
> The only errors in the CI tests that I can find are `ERROR! Invalid callback for stdout specified: debug`. Is there an issue with my feature that needs resolved?

Sorry - CI is still having some issues - working on it

richm:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

vrindle:
LGTM - @richm?

tests/tests_service.yml
```diff
@@ -0,0 +1,212 @@
+- name: Test firewalld zones
```
richm, line 1: Test firewall user defined services

tests/tests_service.yml
```diff
@@ -0,0 +1,212 @@
...
+  hosts: all
+  become: true
+  tasks:
+    - name: test firewalld zones
```
richm, line 5: test firewall user defined services

BrennanPaciorek:
Oh, one thing I should ask, should I provide more instruction in how to use the feature in the role's README file?

richm (commented):
One more thing - we need this documented in the README.md file - especially the behavior concerning the use of `state` with service - that is, if `newservice` already exists, then
```yaml
service: newservice
state: absent
```
will remove the service, but if you use
```yaml
service: newservice
protocol: http
state: absent
```
it will just remove the `protocol` from the service

tests/tests_service.yml
```diff
@@ -0,0 +1,212 @@
...
+          vars:
+            firewall:
+              - service: http
+                permanent: yes
+                state: present
+              - service: systemroletest
+                permanent: yes
+                state: absent
+
+        - name: Fail if clean state did not work
```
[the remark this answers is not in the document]
vrindle: This would be with `firewall-cmd --query-service` to see if it was added or not

tests/tests_service.yml
BrennanPaciorek: I actually typed this one up, so I'm not sure how I managed this one

tests/tests_service.yml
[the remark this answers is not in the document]
BrennanPaciorek: This command would check if the service specified is in the default zone.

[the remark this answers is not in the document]
vrindle: `firewall-cmd --permanent --get-services` prints predefined services as a space separated list.
`firewall-cmd --query-service=service` will tell you if a service is successfully added or not.

BrennanPaciorek:
working on adding comments to `test_service.yml` so that it is more readable

tests/tests_service.yml
```diff
@@ -0,0 +1,293 @@
...
+                state: absent
+
+        - name: Fail if second removal changes anything
+          fail:
+            msg: something changed when systemrole test removed
+          when: firewall_lib_result.changed
+
+      always:
+        - name: Cleanup
+          tags: tests::cleanup
```
richm, line 288: This should be a list
```yaml
           tags:
             - tests::cleanup
```

tests/tests_service.yml
```diff
@@ -0,0 +1,313 @@
...
+        # Verify service states are as intended
+
+        - name: get all services
+          command: firewall-cmd --permanent --get-services
+          register: result
+
+        - name: Fail if http not present after reset
+          fail:
+            msg: Reset default resulted in unexpected state
+          when: result.stdout.find("http ") == -1
```
richm, line 26: This will fail if the list looks like `ldap ftp http`

I suggest using `firewall-cmd --permanent --query-service=<name>` rather than parsing the stdout of `--get-services` e.g.

```yaml

        - name: Fail if http not present after reset
          command: firewall-cmd --permanent --query-service=http
```

tests/tests_service.yml
```diff
@@ -0,0 +1,313 @@
...
+
+        - name: Fail if systemroletest present prior to test run
+          fail:
+            msg: test service present after reseting defaults
+          when: result.stdout.find("systemroletest") != -1
```
richm, line 31: you can use `command` here too, but use `register` and `failed_when` 
```yaml
        - name: Fail if http not present after reset
          command: firewall-cmd --permanent --query-service=systemroletest
          register:  __result
          failed_when: __result is success
```

tests/tests_service.yml
BrennanPaciorek: I've tried this, it only fetches from the list of services in the default zone. I'll think of a better way, but so far that will not work. Further down I introduced a new test to display the functionality of the command while also using it to test how the user defined services work with zone.

tests/tests_service.yml
BrennanPaciorek: I'll go and update this, thank you.

tests/tests_service.yml
BrennanPaciorek: Here is a screenshot to illustrate https://drive.google.com/file/d/1OVHSG7l1PdPfmQ21F3QPofqJU-HEGXwI/view?usp=sharing

richm: hmm - ok - if you really need to use `--get-services`, then you'll have to use a regex to parse the stdout.

https://docs.ansible.com/ansible/2.9/user_guide/playbooks_tests.html

The pattern has to match something like `(^|\s)theservicename(\s|$)` - beginning of string or space, followed by the name, followed by space or end of string - something like that

richm (approved):
lgtm

richm:
@vrindle ok?

PR #83, opened 2022-06-09
Ansible facts with firewalld configuration

BrennanPaciorek:


BrennanPaciorek:
This should not pass any of the tests yet. This is the earliest not broken version of the feature.

BrennanPaciorek:
A couple things that have yet to be added

- firewall_lib_config.py ansible module boilerplate
- ^ https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general.html#developing-modules-general (dropping this here so i can close a tab)
- possibly a pre-[configure firewall] play to configure the ansible fact before attempting to run anything else, in case someone wants to view configuration if their configuration fails
- python 2.7 compatibility (can use os library or pathlib2 - https://pypi.org/project/pathlib2/)
- more details to the ansible fact (?)
- compatibility with CI tests otherwise

richm:
fyi - here is an example of a facts module from the selinux role - https://github.com/linux-system-roles/selinux/blob/master/library/selinux_modules_facts.py
and here is an example of one from the timesync role that is a shell script (Ansible modules do not have to be python): https://github.com/linux-system-roles/timesync/blob/master/library/timesync_provider.sh

vrindle:
The structure of the files and the work you've done so far looks good. It definitely needs testing, but it does look good so far :D

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,116 @@
...
+from ansible.module_utils.basic import AnsibleModule
+import os
+
+
+def config_to_dict(module):
+    config = {}
+    defaults = {}
+    custom = {}
+
+    defaults_path = os.path.join("/", "usr", "lib", "firewalld")
```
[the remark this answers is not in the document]
vrindle: https://stackoverflow.com/questions/16010992/how-to-use-directory-separator-in-both-linux-and-windows-in-python/16011057#16011057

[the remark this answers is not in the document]
BrennanPaciorek: Is there any purpose in doing so? This role should run exclusively on Linux, so the explicit "/" does not seem like it would be an issue

Although, looking at this, I realized that os.path.join() is probably unfit for what I was trying to achieve here, as I could probably just write out the string "/usr/lib/firewalld"

BrennanPaciorek:
There are two ways of going about parsing information for each option (service, zone, etc.), you can directly parse the xml for each one, or you can get the information using the firewalld client. I personally prefer the first one, as the code for doing so will be arguably cleaner with less conditionals for each setting.

library/firewall_lib_facts.py
```diff
@@ -78,9 +78,16 @@ def config_to_dict(module, fw=None):
     defaults = {}
     custom = {}
 
-    for setting in ["zones", "services", "policies", "icmptypes", "helpers", "ipsets"]:
-        default_setting_dir = os.path.join(USR_LIB_FIREWALLD, setting)
-        custom_setting_dir = os.path.join(ETC_FIREWALLD, setting)
+    if firewall.config.VERSION >= "0.9.0":
```
richm, line 81: I'm not sure if this is a robust way of doing version comparison - usually `LooseVersion` is used - but at any rate, the more Pythonic way to do this would be a method similar to "duck typing" - see if there is a policies config in the `firewall.config` module e.g.
```
try:
    firewall.config.FIREWALLD_POLICIES
    HAS_POLICIES = true
except NameError:
    HAS_POLICIES = false
...
```

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def config_to_dict(module, fw=None):
+    config = {}
+    defaults = {}
+    custom = {}
+
+    try:
+        firewall.config.FIREWALLD_POLICIES
+        HAS_POLICIES = True
```
richm, line 83:
```python
    try:
        firewall.config.FIREWALLD_POLICIES
        setting_list.append("policies")
    except AttributeError:
        pass
```

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,147 @@
...
+        setting_list = [
+            "zones",
+            "services",
+            "policies",
+            "icmptypes",
+            "helpers",
+            "ipsets",
+        ]
+    else:
+        setting_list = ["zones", "services", "icmptypes", "helpers", "ipsets"]
```
richm, line 97: then you can get rid of these `if HAS_POLICIES/else` clauses

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,136 @@
...
+
+    module = AnsibleModule(argument_spec={}, supports_check_mode=True)
+
+    if not HAS_FIREWALLD:
+        module.fail_json(msg="firewalld not installed")
+
+    if module.check_mode:
+        module.exit_json(**results)
+
+    results["ansible_facts"]["firewalld_config"] = config_to_dict(module)
```
richm, line 128: Why is this being returned as `ansible_facts.firewalld_config` then you have to do
```yaml
  set_fact:
    firewall_config: "{{ __firewalld_facts.ansible_facts.firewalld_config }}"
```
to set `firewall_config`?  I guess you could shorten it by dropping the `ansible_facts` part

BrennanPaciorek: Yeah, I added the ansible facts to account for any additional ansible facts that might be added later on, although I cannot think of any more that would be used for this role, so I can drop the ansible_facts part.

tasks/main.yml
```diff
@@ -67,12 +67,14 @@
   register: firewall_lib_result
 
 - name: gather firewalld configuration
-  when: not firewall | bool
+  when: not firewall | bool and not __firewall_previous_replaced | bool
```
richm, line 70: Why do you need to check for `__firewall_previous_replaced`?

tests/tests_firewall_fact.yml
```diff
@@ -0,0 +1,86 @@
...
+            msg: Ansible fact firewall_config is not defined
+          when: firewall_config is undefined
+
+        - name: Fail if firewall_config custom has attributes
+          vars:
+            custom_item: "{{ firewall_config.custom[item] | list }}"
+          fail:
+            msg: "{{ custom_item }}"
+          when: custom_item | length != 0
+          loop: "{{ firewall_config.custom.keys() | list }}"
```
richm, line 31: I think you could just check for the length of custom:
```
        - name: Fail if firewall_config custom has attributes
          fail:
            msg: There should be no custom config
          when: firewall_config.custom | length > 0
```
This assumes that `firewall_config.custom` is always defined (I think it is).  The `length` of a `dict` is the number of keys, so an empty `dict` should have `length == 0`

tests/tests_firewall_fact.yml
```diff
@@ -0,0 +1,86 @@
...
+
+        - name: Fail if firewall_config defaults do not have attributes
+          vars:
+            default_item: "{{ firewall_config.default[item] | list }}"
+          fail:
+            msg: default {{ item }} does not have attributes
+          when: default_item | length == 0 and item != "ipsets"
+          loop: "{{ firewall_config.default.keys() | list }}"
```
richm, line 39: same as above, except `when: firewall_config.default | length == 0`

tests/tests_firewall_fact.yml
```diff
@@ -0,0 +1,86 @@
...
+        - name: save default ansible fact value
+          set_fact:
+            __previous_firewall_config: "{{ firewall_config }}"
+          no_log: true
+
+        - name: modify firewalld configuration
+          shell:
+            cmd: |
+              firewall-cmd --permanent --add-service https
+              firewall-cmd --permanent --new-service custom
```
richm, line 50: I think one of these is supposed to be a zone?  And does `https` exist by default?

tests/tests_firewall_fact.yml
```diff
@@ -0,0 +1,86 @@
...
+          fail:
+            msg: firewall customs did not change (when they should have)
+          when: >-
+            firewall_config.custom.zones
+            == __previous_firewall_config.custom.zones
+            and firewall_config.custom.services
+            == __previous_firewall_config.custom.services
+
+      always:
+        - name: Cleanup
```
richm, line 80: please add
```yaml
          tags:
            - tests::cleanup
```
to this task - we sometimes skip cleanup (using `ansible-playbook ... --skip-tags tests::cleanup ...`) in some test scenarios

tests/tests_firewall_fact.yml
BrennanPaciorek: those are just to show that modified defaults present themselves in custom dictionary. The first one is to add https to the default zone's permanent configuration

tests/tests_firewall_fact.yml
```diff
@@ -78,6 +78,7 @@
 
       always:
         - name: Cleanup
+          tags: tests::cleanup
```
richm, line 81: this should a list - so
```yaml
           tags:
             - tests::cleanup
```

BrennanPaciorek: Oh, my mistake. This has been fixed now on both PRs.

tasks/main.yml
BrennanPaciorek: Because it collects the fact if you specify only previous replaced, the prior loop strips `previous` from firewall from what I've noticed when writing the test

tests/tests_firewall_fact.yml
BrennanPaciorek: Yes firewall_config.custom is always defined, but so are all of the options, so in any case the length of firewall_config.custom would be 5 or 6

tasks/main.yml
richm: hmm - if that loop is modifying the `firewall` variable in place, that's really bad - can you put a
```yaml
- name: show firewall
  debug:
    msg: firewall is {{ firewall | to_nice_json }}
```
just after the call to `firewall_lib`?

BrennanPaciorek: Looks like I was incorrect, firewalld is not altered like i thought it was. My mistake

BrennanPaciorek: Yeah, blocked the two ansible fact plays into one, then placed the conditional on the block, and that fixed the problem I was having, pushing changes up now after removing the debug plays.

README.md
```diff
@@ -27,6 +27,82 @@ WARNING: If the configuration failed or if the firewall configuration limits
...
+- `icmptypes`
+- `helpers`
+- `ipsets`
+- `policies` (if supported by remote host's firewalld installation)
+
+Each of the keys contains a list of elements present in 
+permanent configuration for each respective option.
+
+```json
+firewall_config: {
```
richm, line 61: `"firewall_config": {` in JSON

tasks/main.yml
```diff
@@ -66,6 +66,18 @@
...
+  block:
+    - name: gather firewall config information
+      firewall_lib_facts:
+      register: __firewalld_facts
+
+    - name: update firewalld_config fact
+      set_fact:
+        firewall_config: >-
+          "{{ __firewalld_facts.ansible_facts.firewalld_config }}"
+  when: firewall == None
```
richm, line 79: I think you should use the boolean here - `when: not firewall | bool` - just in case someone tries `firewall: {}` or `firewall: []`

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,136 @@
...
+
+    return config
+
+
+def fetch_settings_from_dir(directory):
+    setting_options = []
+
+    for _file in os.listdir(directory):
+        is_setting = ".xml" in _file and ".old" not in _file
+        if is_setting:
```
richm, line 106: you could omit `is_setting` and just use
```python
         if _file.endswith(".xml"):
```

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,136 @@
...
+        setting_list.append("policies")
+    except AttributeError:
+        pass
+
+    for setting in setting_list:
+        default_setting_dir = os.path.join(firewall.config.USR_LIB_FIREWALLD, setting)
+        custom_setting_dir = os.path.join(firewall.config.ETC_FIREWALLD, setting)
+
+        defaults[setting] = fetch_settings_from_dir(default_setting_dir)
+        custom[setting] = fetch_settings_from_dir(custom_setting_dir)
```
richm, line 93: To make it easier for users to test for empty settings, I suggest this:
```python
        settings = fetch_settings_from_dir(default_setting_dir)
        if settings:
            defaults[setting] = settings
        settings = fetch_settings_from_dir(custom_setting_dir)
        if settings:
            custom[setting] = settings
```

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,136 @@
...
+
+    config["default"] = defaults
```
richm, line 95:
```python
     if defaults:
         config["default"] = defaults
     if custom:
         config["custom"] = custom
````

File: tests/tests_firewall_fact.yml (hunk 7 above)
Comment (richm) [reply tests/tests_firewall_fact.yml:31]: ok - if you make the changes I suggested in the python code, the emptiness check should be easier, but you'll have to check if `when: "'custom' in firewall_config"` or similar

File: tasks/main.yml (hunk 13 above)
Comment (BrennanPaciorek) [reply tasks/main.yml:79]: Yeah that is what I started with, but `firewall = [{previous: replaced}]` makes it through the conditional `not firewall | bool` for no good reason.

Comment (BrennanPaciorek) [reply tasks/main.yml:79]: Note: I don't think it is only `previous` that causes this issue, but that is what I use to test the conditional at the moment

Comment (richm) [reply tasks/main.yml:79]: ok - yes - I guess in this particular context, a non-empty dict evaluates to boolean `False` for some reason
then use
```yaml
  when: firewall is none or firewall | length == 0
```
note that `when` (and other conditionals such as `that`, `failed_when`, `changed_when`, etc.) are evaluated as Jinja expressions (that is, you don't have to use `{{ ... }}`) - use lowercase `none` - https://jinja.palletsprojects.com/en/2.9.x/templates/#literals (not sure if this is true for jinja 2.7, but I believe it works)

BrennanPaciorek:
Added features discussed, ran into a sanity test issue where the test cannot import firewall, does anyone know why this is showing for firewall_lib_facts.py but not firewall_lib.py? I do not see any ignores on this particular test. @vrindle @richm

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,204 @@
...
+except ImportError:
+    HAS_FIREWALLD = False
+
+try:
+    if HAS_FIREWALLD:
+        firewall.config.FIREWALLD_POLICIES
+
+    HAS_POLICIES = True
+except AttributeError:
+    HAS_Policies = False
```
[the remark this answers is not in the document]
BrennanPaciorek: Thanks

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def config_to_dict(module, fw):
+    detailed = module.params["detailed"]
```
richm, line 96:
```python
  detailed = module.params.get("detailed", False)
```

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,204 @@
...
+    setting_list = ["zones", "services", "icmptypes", "helpers", "ipsets"]
+
+    if HAS_POLICIES:
+        setting_list.append("policies")
+
+    for setting in setting_list:
+        default_setting_dir = os.path.join(firewall.config.USR_LIB_FIREWALLD, setting)
+        custom_setting_dir = os.path.join(firewall.config.ETC_FIREWALLD, setting)
+
+        settings = (
```
richm, line 109:
```python
  settings = fetch_settings_from_dir(default_setting_dir, detailed, fw)
```

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,204 @@
...
+
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=True)
+
+    if not HAS_FIREWALLD:
+        module.fail_json(msg="firewalld not installed")
+
+    if module.check_mode:
+        module.exit_json(**results)
+
+    fw = FirewallClient()
```
richm, line 194: If you don't use `fw` in this context, maybe declare it in `config_to_dict` or `fetch_settings_from_dir`, and only if you need it?  e.g.
```python
  fw = None
  if I_need_fw:
      fw = FirewallClient()
```

tasks/main.yml
```diff
@@ -63,14 +63,63 @@
...
         value: replaced
+    __detailed:
+      - key: detailed
+        value: true
+      - key: detailed
+        value: false
+      - key: detailed
+        value: yes
+      - key: detailed
+        value: no
```
richm, line 81: true and yes should both be evaluated into `True` by the yaml parser, same with false and no into `False`.  If not, then you might need to add sections for values `on` and `off` as well

tasks/main.yml
```diff
@@ -63,14 +63,63 @@
...
   register: firewall_lib_result
 
+- name: gather firewall ansible fact with args
+  vars:
+    detailed: "detailed"
```
richm, line 86: you should not use double quotes for simple strings

tasks/main.yml
BrennanPaciorek: You're probably right, I didn't try just `True` and `False` values yet

tests/tests_service.yml
```diff
@@ -312,4 +312,4 @@
             name: linux-system-roles.firewall
           vars:
             firewall:
-              previous: replaced
+              - previous: replaced
```
richm, line 315: hmm - you may have found a bug in https://github.com/linux-system-roles/firewall/blob/master/tasks/main.yml#L65 - if the dict form is not working - I believe this may need to be 
```yaml
  loop: "{{ (firewall is mapping) | ternary([firewall], firewall) |
```
That is - the `|` takes precedence over the `is`

BrennanPaciorek: This fails before it reaches the loop, I'll fix the cause but it is the when for that block of plays causing dict form not to fail, so its a bug I just created. I'll fix it.

BrennanPaciorek:
Looks like getSettingsDict() is not available on the latest supported version of firewalld for RHEL 7. working on fixing that now

tasks/main.yml
BrennanPaciorek: removed the yes/no key-value pairs, and there are no issues when running the role with detailed set

tests/tests_service.yml
BrennanPaciorek: fixed. Left a dict form `previous: replaced` to make sure there is some coverage of dict form in the test suite now

README.md
```diff
@@ -27,6 +27,118 @@ WARNING: If the configuration failed or if the firewall configuration limits
...
+"default": {
+  "zones": [{}...],
+  "services": [{}...],
+  "icmptypes": [{}...],
+  "helpers": [{}...],
+  "ipsets": [{}...],
+  "policies": [{}...],
+}
+```
+
```
BrennanPaciorek, line 112: Is this enough to accurately represent the structure of the ansible fact? I definitely cannot provide a full example for a detailed firewall_config.default, because I am certain that approaches 10,000 lines if shown in full.

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,220 @@
...
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
BrennanPaciorek, line 20: Does this heading look correct? It feels incorrect to list myself as the sole author here

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,220 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = """
+module: firewall_lib_facts
+short_description: module for firewall role
+requirements:
+  - python3-firewall or python-firewall
+description:
+  Get firewalld configuration information, including default settings
+  and custom settings. There are no options for this module
```
BrennanPaciorek, line 33: There are options now, need to update description to match

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,220 @@
...
+          "default": {},
+          "custom": {},
+      },
+  }
+"""
+
+EXAMPLES = """
+# Run with no parameters to gather facts
+firewall_lib_facts:
+"""
```
BrennanPaciorek, line 71: is EXAMPLES necessary? The user will be accessing this module via the firewall var.

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,220 @@
...
+                element = fw.config().getServiceByName(_item).getSettings()
+                try:
+                    element_settings = element.getSettingsDict()
+                except AttributeError:
+                    element_settings["version"] = element.getVersion()
+                    element_settings["short"] = element.getShort()
+                    element_settings["description"] = element.getDescription()
+                    element_settings["protocols"] = element.getProtocols()
+                    element_settings["source_ports"] = element.getSourcePorts()
+                    element_settings["helpers"] = element.getModules()
```
BrennanPaciorek, line 169: This actually fetches modules, so I'll change this to `element_settings["modules"]` to also better represent the alternative dictionary `getSettingsDict()` provides.

BrennanPaciorek (commented):
The review I made is to actively seek feedback on those points and to organize my thoughts. The ones without questions are more of a reminder than anything, but if you object to any changes I suggested of myself, feel free to respond to those conversations as well.

README.md
richm: Right.  Just enough to give the reader some rough idea about what it will look like.

Really?  10,000 lines?  Then I'm wondering how a user will actually use this at all, much less print it - something like `debug: msg: firewall config {{ firewall_config | to_nice_json }}` is likely to blow up the ansible log

library/firewall_lib_facts.py
richm: Looks good to me.

README.md
BrennanPaciorek: It does, thats why i supress logs and why i make `detailed: false` the default

library/firewall_lib_facts.py
richm: There is no problem listing yourself as sole author.

README.md
BrennanPaciorek: just checked ansible log, 10000 was a little high, ~4600 is a lot more accurate

tasks/main.yml
```diff
@@ -63,14 +63,59 @@
...
+      register: __firewalld_facts
+
+    - name: update firewalld_config fact
+      vars:
+        firewalld_facts: "{{ __firewalld_facts.results[0] }}"
+      set_fact:
+        firewall_config: "{{ firewalld_facts.firewall_config }}"
+  when: >-
+    firewall != None and firewall | length == 1 and
+    detailed in fw[0].keys()
```
richm, line 103: if `fw[0]` is a `dict` then you should be able to use `detailed in fw[0]`

README.md
```diff
@@ -252,7 +364,7 @@ list of destination addresses, option only implemented for user defined services
 Takes 0-2 addresses, allowing for one IPv4 address and one IPv6 address or address range.
 
 IPv4 format: x.x.x.x[/mask]
-IPv6 format: x:x:x:x:x:x:x:x[/mask] (x::x works when abbreviating one or more subsequent IPv6 segments where x = 0)
+IPv6 format: x\:x\:x\:x\:x\:x\:x\:x[/mask] (x::x works when abbreviating one or more subsequent IPv6 segments where x = 0)
```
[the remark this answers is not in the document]
BrennanPaciorek: I can switch it to code formatting, the issue was that `:x:` created this : :x:

tasks/main.yml
richm: The ansible recommendation for multiple `and` clauses is to use the multi-line form of `when`:
```yaml
  when:
    - not firewall is none
    - firewall | length == 1
    - detailed in fw[0]
```

tests/tests_firewall_fact.yml
```diff
@@ -0,0 +1,128 @@
...
+          fail:
+            msg: Ansible fact firewall_config is not defined
+          when: firewall_config is undefined
+
+        - name: Fail if firewall_config custom exists
+          vars:
+            custom: custom
+          fail:
+            msg: custom should not exist if there is no custom configuration
+          when: custom in firewall_config.keys()
```
richm, line 32: `custom in firewall_config`

Also - you can use this form for `when` if the first thing in the `when` is a quoted string:
```
  when: "'custom' in firewall_config"
```
if you want to avoid the `custom` var

tests/tests_firewall_fact.yml
```diff
@@ -0,0 +1,128 @@
...
+            __default_config_iterable: "{{ __default_config_iterable +  item }}"
+          loop: "{{ firewall_config.default.values() | list }}"
+
+        - name: Fail if default is not list of dictionaries
+          fail:
+            msg: >-
+              detailed firewall_config.default not formatted properly -
+              see README.md for expected formatting
+          when: item is not mapping
+          loop: "{{ __default_config_iterable }}"
```
richm, line 112: can't you use `loop: "{{ firewall_config.default.values() }}"` and avoid the `set_fact` tasks?

richm: hmm - so the issue is that when you use `loop: "{{ firewall_config.default.values() }}"`, each `item` looks like this:
```
- key: value
  key1: value1
...
```
and you want `item` to look like this:
```
key: value
key1: value1
```
?

If so, can you use the `flatten` filter?  https://docs.ansible.com/ansible/2.9/user_guide/playbooks_filters.html#list-filters
`loop: "{{ firewall_config.default.values() | flatten }}"`

BrennanPaciorek: Oh, that would work, thank you.

README.md
```diff
@@ -27,6 +27,171 @@ WARNING: If the configuration failed or if the firewall configuration limits
...
+  firewall:
+```
+
+Another option is to gather a more detailed version of the
+ansible facts by using the detailed argument e.g.
+```yaml
+vars:
+  firewall:
+    detailed: true
+```
```
BrennanPaciorek, line 48: Add a flashy warning about th size of the fact when `detailed`?

tasks/main.yml
```diff
@@ -63,14 +63,61 @@
...
+      loop: "{{ fw | map('dict2items') | map('difference', __previous) |
+        select | map('items2dict') | list }}"
+      register: __firewalld_facts
+
+    - name: update firewalld_config fact
+      vars:
+        firewalld_facts: "{{ __firewalld_facts.results[0] }}"
+      set_fact:
+        cacheable: yes
+        firewall_config: "{{ firewalld_facts.firewall_config }}"
```
BrennanPaciorek, line 100: New parameter for setting cacheable? This `set_fact` argument allows for permanent storage of the fact, but it also causes the variable to be created twice in two different scopes. If used with the default plugin (builtin.memory) it doesn't really do all that much besides use up the extra RAM.

tasks/main.yml
```diff
@@ -63,14 +63,61 @@
...
+  block:
+    - name: gather firewall config if no arguments
+      firewall_lib_facts:
+        detailed: false
+      register: __firewalld_facts
+
+    - name: update firewalld_config fact
+      set_fact:
+        cacheable: yes
+        firewall_config: "{{ __firewalld_facts.firewall_config }}"
```
BrennanPaciorek, line 116: same concern as above

library/firewall_lib_facts.py
```diff
@@ -0,0 +1,221 @@
...
+
+    HAS_POLICIES = True
+except AttributeError:
+    HAS_POLICIES = False
+
+
+def config_to_dict(module):
+    detailed = module.params.get("detailed", False)
+    config = {}
+    defaults = {}
```
BrennanPaciorek, line 98: Remove default sub-dictionary? I personally think we should keep it for better compatibility with use cases in the email. Aside from that, all we would need to do is decorate the README with warnings, both about cacheable (see tasks/main.yml review) and the size of `detailed` firewall_config ansible fact.

BrennanPaciorek (commented):
@richm @vrindle

README.md
richm: Yes - something like `WARNING: The detailed firewall facts may be approximately 1Mb in size` or whatever the size is.

library/firewall_lib_facts.py
richm: I think we should keep it.

tasks/main.yml
richm: Can't we just leave this up to the user to configure the caching outside of the role?  Do we have to indicate `cacheable` here?

richm:
This lgtm - @vrindle ?

vrindle:
@richm lgtm

PR #84, opened 2022-06-22
make all tests work with gather_facts: false

richm:
The tests_zone.yml test uses facts outside of the role and
needs to `gather_facts: true` when using ANSIBLE_GATHERING=explicit


nhosoi (approved):
lgtm

Issue #85, opened 2022-06-23
forward_port should accept list of string or list of dict


BrennanPaciorek:
**What happened**:
`forward_port` argument validation does not accept dictionary argument
**What you expected to happen**:
`forward_port` functionality to work when using a dictionary argument instead of a string, which is the deprecated format
**How to reproduce it (as minimally and precisely as possible)**:
```yaml
firewall:
  forward_port:
    - port: [any port]
      proto: [tcp/udp]
```
**Anything else we need to know?**:
No

PR #86, opened 2022-06-23
bugfix: port forward dict form works

BrennanPaciorek:
- fixed bug where port_forward argument only worked with string argument

 - additionally argument converts to list if necessary

- minimal tests added for port forward

 - tests_port_forward.yml only has the fail case that the role fails

Fixes: #85

### DIFF
--- a/library/firewall_lib.py
+++ b/library/firewall_lib.py
@@ -65,12 +65,13 @@ options:
   forward_port:
     description:
       List of forward port strings.
+      Elements must be str or dict.
       The format of a forward port needs to be
       <port>[-<port>]/<protocol>;[<to-port>];[<to-addr>].
     aliases: ["port_forward"]
     required: false
     type: list
-    elements: str
+    elements: raw
   masquerade:
     description:
       The masquerade bool setting.
@@ -390,7 +391,7 @@ def parse_forward_port(module, item):
         else:
             _to_port = None
         _to_addr = item.get("toaddr")
-    else:
+    elif isinstance(item, str):
         args = item.split(";")
         if len(args) == 3:
             __port, _to_port, _to_addr = args
@@ -404,6 +405,10 @@ def parse_forward_port(module, item):
             _to_port = None
         if _to_addr == "":
             _to_addr = None
+    else:
+        module.fail_json(
+            msg="improper %s type (must be str or dict): %s" % (type_string, item)
+        )
 
     return (_port, _protocol, _to_port, _to_addr)
 
@@ -421,7 +426,7 @@ def main():
             forward_port=dict(
                 required=False,
                 type="list",
-                elements="str",
+                elements="raw",
                 default=[],
                 aliases=["port_forward"],
                 deprecated_aliases=[
--- a/library/firewall_lib.py
+++ b/library/firewall_lib.py
@@ -64,14 +64,13 @@ options:
     elements: str
   forward_port:
     description:
-      List of forward port strings.
-      Elements must be str or dict.
-      The format of a forward port needs to be
+      List of forward port strings or dicts,
+      or a single string or dict.
+      The format of a forward port string needs to be
       <port>[-<port>]/<protocol>;[<to-port>];[<to-addr>].
     aliases: ["port_forward"]
     required: false
-    type: list
-    elements: raw
+    type: raw
   masquerade:
     description:
       The masquerade bool setting.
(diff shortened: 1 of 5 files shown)


tasks/main.yml
```diff
@@ -44,7 +44,8 @@
     service: "{{ item.service | default(omit) }}"
     port: "{{ item.port | default(omit) }}"
     source_port: "{{ item.source_port | default(omit) }}"
-    forward_port: "{{ item.forward_port | default(omit) }}"
+    forward_port: "{{ (item.forward_port is mapping) |
+      ternary([item.forward_port], item.forward_port) | default(omit) }}"
```
richm, line 48: I don't think you need to change this - the type of `item.forward_port` should always be a `list` https://github.com/linux-system-roles/firewall/blob/6521366e9ef72492ff7563fb5c5073ea375796d1/library/firewall_lib.py#L428 and

tests/tests_ansible.yml
```diff
@@ -525,6 +525,60 @@
...
+        failed_when: result is failed or result is not changed
+
+      - name: forward same port again (string)
+        firewall_lib:
+          forward_port: 40/tcp;8080;0.0.0.0
+          state: enabled
+        register: result
+        failed_when: result is failed or result is changed
+
+      - name: diable forward port (string)
```
richm, line 542: disable

tasks/main.yml
BrennanPaciorek: Oh, should we not allow support for something like:
```yaml
firewall:
  forward_port:
    port: 80
    proto: tcp
```
(When "-" is not placed before the port being forwarded)

richm: ok - I see - https://github.com/linux-system-roles/firewall#forward_port "dict or list of dicts in the format like ansible.posix.firewalld"
Then I think the type handling should be done in the module, not here - I think the logic is going to get complicated, because the value of `forward_port` could be a single string, a list of strings, a single dict, or a list of dict - or even a list of mixed dict and string
But that means we'll have to change https://github.com/linux-system-roles/firewall/blob/6521366e9ef72492ff7563fb5c5073ea375796d1/library/firewall_lib.py#L428 to also be `raw` . . . on the other hand, every other similar parameter says "str or list" in the docs e.g. https://github.com/linux-system-roles/firewall#port - but they should all fail Ansible argument validation because they all take a `list` e.g. https://github.com/linux-system-roles/firewall/blob/6521366e9ef72492ff7563fb5c5073ea375796d1/library/firewall_lib.py#L424 - I'm assuming if you try to pass a string instead of a list, you will get an error - so really the documentation is wrong, or the doc writer intended to support both `str` and `list` (somehow - by writing custom Ansible argument validation code?)

I guess what I'm saying is that we should just assume the value of `forward_port` here is a `list`

Then, I guess we can have a separate discussion about what to do about the README
* Change it to reflect the actual code - only `list` is accepted
* Change the code to reflect the README - which means implementing our own argument validation

tests/tests_forward_port.yml
```diff
@@ -0,0 +1,44 @@
...
+          - forward_port: 40/tcp;8080;0.0.0.0
+            state: disabled
+
+    - name: forward_port (dict) and undo
+      include_role:
+        name: linux-system-roles.firewall
+      vars:
+        firewall:
+          - forward_port:
+              port: 40
```
richm, line 32: see above

tasks/main.yml
BrennanPaciorek: I believe they support a string value for `port`, they ansible just appears to pass strings in as `[string]`, which works for a single port. The issue is when a `dict` is passed in to a list and ansible-playbook does not seem to know what to do to convert it to a list like it does with `str`

Nonetheless, I do find using our own argument validation to be largely unnecessary if there is no issue with assuming a list is passed in. I'll close this conversation and reopen another over README

BrennanPaciorek: What I'm going to do for the time being is update README to match the actual functionality that is present. This would just mean removing the mention of single dict in the `forward-port` section.

BrennanPaciorek: Update, re-implemented single dict with minimal changes in firewall_lib.py

library/firewall_lib.py
```diff
@@ -64,13 +64,13 @@
     elements: str
   forward_port:
     description:
-      List of forward port strings.
-      The format of a forward port needs to be
+      List of forward port strings or dicts,
+      or a single string or dict.
+      The format of a forward port string needs to be
```
richm, line 69: The string format of . . .

richm: never mind

richm:
lgtm - @vrindle ?

PR #89, opened 2022-06-30
Add/remove interfaces to zone using PCI device ID

BrennanPaciorek:
- Add/remove interfaces to a zone using PCI ID (formatted XXXX:XXXX, VendorID:DeviceID)
  - Collision of interfaces with the same PCI ID (while unlikely) is handled by applying the play to each interface with the PCI id
  - fixes RHBZ 2100939
  - Useful as it allows control nodes to manage interfaces by what a device is rather than what interface name it was assigned.
- Added Network Manger interfacing in the same capacity to which firewall-cmd interacts with Network Manager for interfaces
  - See `--[add/remove/change]-interface`  at https://firewalld.org/documentation/man-pages/firewall-cmd.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,9 +200,62 @@ source: 192.0.2.0/24
 
 String or list of interface name strings.
 
-```
+```yaml
 interface: eth2
 ```
+
+This role handles interface arguments similar to
+how firewalld's cli, `firewall-cmd` does, i.e.
+manages the interface through NetworkManger if possible,
+and handles the interface binding purely through firewalld
+otherwise.
+
+```
+WARNING: Neither firewalld nor this role throw any
+errors if the interface name specified is not
+tied to any existing network interface. This can cause confusion
+when attempting to add an interface via PCI device ID,
+for which you should use the parameter `interface_pci_id`
+instead of the `interface` parameter.
+
+Allow interface named '8086:15d7' in dmz zone
+
+firewall:
+  - zone: dmz
+    interface: 8086:15d7
+    state: enabled
+
+The above will successfully add a nftables/iptables rule
+for an interface named `8086:15d7`, but no traffic should/will
+ever match to an interface with this name.
+
+TLDR - When using this parameter, please stick only to using
+logical interface names that you know exist on the device to 
+avoid confusing behavior.
+```
+
+### interface_pci_id
+
+String or list of interface PCI device IDs.
+Accepts PCI IDs if the wildcard `XXXX:YYYY` applies
+where:
+- XXXX: Hexadecimal, corresponds to Vendor ID
+- YYYY: Hexadecimal, corresponds to Device ID
+
+```yaml
+# PCI id for Intel Corporation Ethernet Connection
+interface_pci_id: 8086:15d7
+```
+
+Only accepts PCI devices IDs that correspond to a named network interface,
+and converts all PCI device IDs to their respective logical interface names.
+
+If a PCI id corresponds to more than one logical interface name,
+all interfaces with the PCI id specified will have the play applied.
+
+A list of PCI devices with their IDs can be retrieved using `lcpci -nn`.
+For more information on PCI device IDs, see the linux man page at:
+https://man7.org/linux/man-pages/man5/pci.ids.5.html
 
 ### icmp_block
 
@@ -418,6 +471,14 @@ Do not permit traffic in default zone for https service:
 firewall:
   - service: https
     state: disabled
+```
+
+Allow interface with PCI device ID '8086:15d7' in dmz zone
+```yaml
+firewall:
+  - zone: dmz
+    interface_pci_id: 8086:15d7
+    state: enabled
 ```
 
 Example Playbooks
--- a/library/firewall_lib.py
+++ b/library/firewall_lib.py
@@ -98,6 +98,13 @@ options:
     required: false
     type: list
     elements: str
+  interface_pci_id:
+    description:
+      List of inteface PCI device ID strings.
+      PCI device ID needs to correspond to a named network interface.
+    required: false
+    type: list
+    elements: str
   icmp_block:
     description:
       List of ICMP type strings to block.
@@ -205,6 +212,8 @@ options:
 
 from distutils.version import LooseVersion
 from ansible.module_utils.basic import AnsibleModule
+import re
+import os
 
 try:
     import firewall.config
@@ -221,6 +230,52 @@ try:
     HAS_FIREWALLD = True
 except ImportError:
     HAS_FIREWALLD = False
+
+try:
+    from firewall.core.fw_nm import (
+        nm_is_imported,
+        nm_get_connection_of_interface,
+        nm_get_zone_of_connection,
+        nm_set_zone_of_connection,
+        nm_get_interfaces,
+        nm_get_client,
+    )
+
+    NM_IMPORTED = nm_is_imported()
+except ImportError:
+    NM_IMPORTED = False
+
+
+PCI_REGEX = re.compile("[0-9a-fA-F]{4}:[0-9a-fA-F]{4}")
+
+
+def try_get_connection_of_interface(interface):
+    try:
+        return nm_get_connection_of_interface(interface)
+    except Exception:
+        return None
+
+
+def try_set_zone_of_interface(module, _zone, interface):
+    if NM_IMPORTED:
+        connection = try_get_connection_of_interface(interface)
+        if connection is not None:
+            if _zone == "":
+                zone_string = "the default zone"
+            else:
+                zone_string = _zone
+            if _zone == nm_get_zone_of_connection(connection):
+                module.log(
+                    msg="The interface is under control of NetworkManager and already bound to '%s'"
+                    % zone_string
+                )
+            elif not module.check_mode:
+                nm_set_zone_of_connection(_zone, connection)
+            return True
+    return False
+
+
+# Above: adapted from firewall-cmd source code
 
 
 def create_service(module, fw, service):
@@ -267,6 +322,46 @@ def handle_interface_permanent(
                 old_zone_settings.removeInterface(item)
                 old_zone_obj.update(old_zone_settings)
             fw_settings.addInterface(item)
+
+
+pci_ids = None
+
+
+def get_interface_pci():
+    pci_dict = {}
+    for interface in nm_get_interfaces():
+        # udi/device/[vendor, device]
+        interface_ids = []
+        device_udi = nm_get_client().get_device_by_iface(interface).get_udi()
+        device_path = os.path.join(device_udi, "device")
+        for field in ["vendor", "device"]:
+            with open(os.path.join(device_path, field)) as _file:
+                interface_ids.append(_file.readline().strip(" \n")[2:])
+        interface_ids = ":".join(interface_ids)
+        if interface_ids not in pci_dict:
+            pci_dict[interface_ids] = [interface]
+        else:
+            pci_dict[interface_ids].append(interface)
+    return pci_dict
+
+
+def parse_pci_id(module, item):
+    if PCI_REGEX.search(item):
+        global pci_ids
+        if not pci_ids:
+            pci_ids = get_interface_pci()
+
+        interface_name = pci_ids.get(item)
+        if interface_name:
+            return interface_name
+
+        module.warn(msg="No network interfaces found with PCI device ID %s" % item)
+    else:
+        module.fail_json(
+            msg="PCI id %s does not match format: XXXX:XXXX (X = hexadecimal number)"
+            % item
+        )
+    return []
 
 
 def parse_port(module, item):
@@ -447,6 +542,9 @@ def main():
             rich_rule=dict(required=False, type="list", elements="str", default=[]),
             source=dict(required=False, type="list", elements="str", default=[]),
             interface=dict(required=False, type="list", elements="str", default=[]),
+            interface_pci_id=dict(
+                required=False, type="list", elements="str", default=[]
+            ),
             icmp_block=dict(required=False, type="list", elements="str", default=[]),
             icmp_block_inversion=dict(required=False, type="bool", default=None),
             timeout=dict(required=False, type="int", default=0),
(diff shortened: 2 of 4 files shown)


BrennanPaciorek:
I'm still tweaking the test, as there will most likely be issues when dealing with traditional interface names and systemd predictable interface names

library/firewall_lib.py
```diff
@@ -323,28 +311,33 @@ def handle_interface_permanent(
...
         device_path = os.path.join(device_udi, "device")
         for field in ["vendor", "device"]:
             with open(os.path.join(device_path, field)) as _file:
                 interface_ids.append(_file.readline().strip(" \n")[2:])
-
-        pci_dict[":".join(interface_ids)] = interface
+        interface_ids = ":".join(interface_ids)
+        if interface_ids not in pci_dict:
+            pci_dict[interface_ids] = [interface]
+        else:
```
liangwen12year, line 333: Maybe detailed commit message is needed for explaining why this commit is needed.

README.md
```diff
@@ -199,10 +199,18 @@ source: 192.0.2.0/24
 ### interface
 
 String or list of interface name strings.
+Also accepts PCI id's, which is two pairs of hexidecminal numbers (XXXX:XXXX)
```
liangwen12year, line 202: If the regex applies to the PCI id matching ?

README.md
```diff
@@ -199,10 +199,18 @@ source: 192.0.2.0/24
...
-```
+If a PCI id corresponds to more than one interface name,
+all interfaces with the PCI id specified will have the play applied.
+
+```yaml
 interface: eth2
 ```
+```yaml
+# PCI id for Intel Corporation Ethernet Connection
+interface: 8086:15d7
```
liangwen12year, line 212: FYI, in the network role, we use a brand new key name (`path`) to specify the PCI address [1], not sure if the firewall role should also have a good key name distinction between the actual interface name and PCI ID.

[1] https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L1755

liangwen12year: Also adding an complete vivid example like the following would be nice:

Allow interface with PCI ID 8086:15d7  in trusted zone:

```yaml
firewall:
  - interface: 8086:15d7
    zone: trusted
    state: enabled
```

library/firewall_lib.py
```diff
@@ -222,6 +224,48 @@
...
+    from firewall.core.fw_nm import (
+        nm_is_imported,
+        nm_get_connection_of_interface,
+        nm_get_zone_of_connection,
+        nm_set_zone_of_connection,
+        nm_get_interfaces,
+        nm_get_client,
+    )
+
+    NM_IMPORTED = nm_is_imported()
```
vrindle, line 237: Is there a benefit to calling this function rather than just setting it to true?

library/firewall_lib.py
```diff
@@ -222,6 +224,48 @@
...
+
+# Below: Adapted from firewall-cmd source code
+def try_set_zone_of_interface(module, _zone, interface):
+    if NM_IMPORTED:
+        try:
+            connection = nm_get_connection_of_interface(interface)
+        except Exception:
+            pass
+        else:
+            if connection is not None:
```
vrindle, line 250: Very nested. This should all probably be in its own function. With so many conditionals it can be hard to follow.

BrennanPaciorek: I'll try to make it more readable

library/firewall_lib.py
BrennanPaciorek: The function actually checks if Network Manager itself is imported in fw_nm.py, which is not checked by importing the functions. If the import fails, we can just assume no, but if it succeeds, the function call here I imagine would be better than calling `nm_is_imported()` for each zone changed.
https://github.com/firewalld/firewalld/blob/master/src/firewall/core/fw_nm.py#L31

README.md
BrennanPaciorek: After thinking about it, I agree on both comments on this conversation, it would be difficult to manage an interface named like a PCI device ID without making the change you suggested or a similar change, although I have no idea why someone would knowingly add an interface with the name in the structure of a PCI id.

Additionally, I will go ahead and add an example to interface using a PCI device ID after making the first suggested change

library/firewall_lib.py
BrennanPaciorek: I updated the PR summary, does that suffice?

liangwen12year: Well, here is a good article about how to write the commit message, https://cbea.ms/git-commit/, you can take a look.

richm: The project has a git commit message section in the contribution guide https://linux-system-roles.github.io/contribute.html
"Write a good commit message"

library/firewall_lib.py
```diff
@@ -269,6 +323,39 @@ def handle_interface_permanent(
...
+
+def parse_pci_id(module, item):
+    if PCI_REGEX.search(item):
+        global pci_ids
+        if not pci_ids:
+            pci_ids = get_interface_pci()
+        interface_name = pci_ids.get(item)
+        if interface_name:
+            return interface_name
+    module.fail_json(msg="No network interfaces found with PCI device ID %s" % item)
```
BrennanPaciorek, line 356: TODO: error if PCI id is invalid, rather than a single fail message in parse_pci_id. Unit test for both errors

BrennanPaciorek (commented):
TODO:
- Integration testing: needs to support case that iptables is the backend, meaning need a good way to get firewalld backend. firewalld.conf seems to not be present on installations of some VMs or is in unusual location.

library/firewall_lib.py
```diff
@@ -269,6 +323,44 @@ def handle_interface_permanent(
...
+
+def parse_pci_id(module, item):
+    if PCI_REGEX.search(item):
+        global pci_ids
+        if not pci_ids:
+            pci_ids = get_interface_pci()
+        interface_name = pci_ids.get(item)
+        if interface_name:
+            return interface_name
+        module.fail_json(msg="No network interfaces found with PCI device ID %s" % item)
```
BrennanPaciorek, line 356: I need to change this to a warning, module shouldn't fail because someone tries to add an interface that happens to not exist on the managed node.

tests/tests_interface_pci.yml
```diff
@@ -0,0 +1,78 @@
...
+        string:"org.fedoraproject.FirewallD1.config"
+        string:"FirewallBackend"
+      ignore_errors: yes
+      register: result
+
+    - name: get backend from result
+      set_fact:
+        firewall_backend:
+          "{{ 'nftables' in result.stdout |
+           ternary('iptables', 'nftables', 'iptables') }}"
```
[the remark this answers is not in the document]
BrennanPaciorek: Oh right, we can't be sure this block would work without calling the firewall role first. The ansible fact PR (https://github.com/linux-system-roles/firewall/pull/83) currently uses up the functionality of calling the role with no options. That PR may need to be revised to preserve the functionality of people just looking to install firewalld dependencies and enable the service remotely.

library/firewall_lib.py
```diff
@@ -447,6 +539,7 @@ def main():
             rich_rule=dict(required=False, type="list", elements="str", default=[]),
             source=dict(required=False, type="list", elements="str", default=[]),
             interface=dict(required=False, type="list", elements="str", default=[]),
+            interface_pci=dict(required=False, type="list", elements="str", default=[]),
```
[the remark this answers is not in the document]
vrindle: `interface_id` is a good name but I think we could be more explicit and call it something like `interface_pci_id`. What do you think? @BrennanPaciorek @liangwen12year

library/firewall_lib.py
vrindle: @richm @liangwen12year I see there's an updated commit message, it looks good. What do you folks think?

library/firewall_lib.py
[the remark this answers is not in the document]
BrennanPaciorek: This sounds good to me, I'll make the changes Monday morning.

richm:
[citest commit:ed1b763a125d77dec38ef56204c30e6fa542e9dc]

library/firewall_lib.py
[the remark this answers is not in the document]
liangwen12year: `interface_pci_id` works for me

library/firewall_lib.py
liangwen12year: The updated commit message lgtm.

richm:
[citest commit:ed1b763a125d77dec38ef56204c30e6fa542e9dc]

BrennanPaciorek:
I need to update the test for this feature, i need to do some checking for dbus errors in case there is no "stdout" when fetching FIREWALL_BACKEND

README.md
```diff
@@ -200,10 +200,33 @@ source: 192.0.2.0/24
...
+```yaml
 interface: eth2
 ```
 
+### interface_pci_id
+
+String or list of interface PCI device IDs.
+Accepts PCI IDs if the wildcard `XXXX:YYYY` applies
+where:
+- XXXX: Hexidecimal, corresponds to Vendor ID
```
richm, line 212: Hexadecimal

library/firewall_lib.py
```diff
@@ -269,6 +324,47 @@ def handle_interface_permanent(
             fw_settings.addInterface(item)
 
 
+pci_ids = None
+
+
+def get_interface_pci():
+    pci_dict = {}
+    interfaces = nm_get_interfaces()
+    for interface in interfaces:
```
richm, line 333: you don't need `interfaces` - `for interface in nm_get_interfaces():`

tests/tests_interface_pci.yml
```diff
@@ -0,0 +1,78 @@
...
+              zone: internal
+              interface_pci_id: 1af4:0001
+              state: enabled
+              permanent: yes
+
+        - name: assert pcid not in nftable ruleset
+          command: nft list ruleset
+          register: result
+          failed_when: "result.rc != 0 or '1af4:0001' in result.stdout"
+          when: nftables_backend
```
richm, line 51: I think you need to use `nftables_backend | bool` here to avoid the "bare variable" problem

tests/tests_interface_pci.yml
```diff
@@ -0,0 +1,78 @@
...
+            firewall:
+              zone: internal
+              interface_pci_id: 1af4:0001
+              state: enabled
+              permanent: yes
+
+        - name: assert pcid not in nftable ruleset
+          command: nft list ruleset
+          register: result
+          failed_when: "result.rc != 0 or '1af4:0001' in result.stdout"
```
richm, line 50: `failed_when: result is failed or '1af4:0001' in result.stdout`

tests/tests_interface_pci.yml
```diff
@@ -0,0 +1,78 @@
...
+          when: nftables_backend
+
+        - name: assert pcid not in iptables rules
+          command: iptables -S
+          register: result
+          failed_when: "result.rc != 0 or '1af4:0001' in result.stdout"
```
richm, line 56: `failed_when: result is failed or '1af4:0001' in result.stdout`

tests/unit/test_firewall_lib.py
```diff
@@ -360,6 +374,24 @@ def test_parse_forward_port(self):
...
+    @patch("firewall_lib.HAS_FIREWALLD", True)
+    @patch("firewall_lib.pci_ids", {"600D:7C1D": ["eth0"]})
+    def test_parse_pci_id(self, am_class):
+        am = am_class.return_value
+
+        am.params = {"interface_pci_id": ["123G:1111"]}
+        with self.assertRaises(MockException):
+            firewall_lib.main()
+        am.fail_json.assert_called_with(
+            msg="PCI id 123G:1111 does not match format: XXXX:XXXX (X = hexidecimal number)"
```
richm, line 387: hexadecimal

library/firewall_lib.py
```diff
@@ -269,6 +324,47 @@ def handle_interface_permanent(
...
+            pci_ids = get_interface_pci()
+
+        interface_name = pci_ids.get(item)
+        if interface_name:
+            return interface_name
+
+        module.warn(msg="No network interfaces found with PCI device ID %s" % item)
+    else:
+        module.fail_json(
+            msg="PCI id %s does not match format: XXXX:XXXX (X = hexidecimal number)"
```
richm, line 362: hexadecimal

BrennanPaciorek:
Okay that should fix the testsuite, CI tests just need to be run to be sure. All feedback has been implemented in the same update to the branch

BrennanPaciorek:
[citest commit:eb1d5b1e41b20cb7cf808ded3e2aedad1653e24f]

richm:
[citest]

richm:
ok - lgtm - @vrindle @liangwen12year ?

README.md
```diff
@@ -420,6 +443,22 @@ firewall:
     state: disabled
 ```
 
+Allow interface named '8086:15d7' in dmz zone
```
briansmith0, line 446: Not sure I understand this example -- are there circumstances where the interface name would be a PCI device ID?

BrennanPaciorek: Firewalld supports adding interfaces by any name, if it doesn't match with any logical interface name, it still gets added as a firewall rule without going through NetworkManager, but it likely will not match with any traffic. Since firewalld supports this, I felt it would be better practice to support it as well, even if it is just for removing interface names that do not match up with any network traffic.

briansmith0: Thanks for the context.  I would suggest that we mention this info you provided under the "interface" section of the README and remove this particular example from the README as it sounds like this example is not something that would generally be used or recommended, and the example might cause some confusion (it did for me at least).  What are your thoughts on this?

BrennanPaciorek: I agree, if it mislead you, I should change how the information is presented. I will move this information over to the interface section as a warning, since it is something the general user would need to know about when adding/moving interfaces.

BrennanPaciorek: Updated README, should the new warning under `interface` be visible enough to users?

briansmith0: This looks good to me and is very easy to understand.  Thanks!

richm:
ok - I think this is ready to merge, unless there are any other objections?

vrindle:
LGTM

PR #98, opened 2022-07-27
better way to get main branch [citest skip]

richm:


richm:
The current version doesn't get the branch name in all cases: https://github.com/linux-system-roles/firewall/runs/7530456114?check_suite_focus=true
```
++ grep -o 'origin/HEAD -> origin/.*$'
++ git branch -r
++ awk -F/ '{print $3}'
+ _branch=
Error: Process completed with exit code 1.
```
This is because of the way that the checkout action works.  If I use the same method as the checkout action:
```
git branch -r
  origin/main
```
so if this fails, try using `git branch --points-at HEAD` - this assumes the code is checked out to the main branch.  If you look at the logs from the checkout action of https://github.com/linux-system-roles/firewall/runs/7530456114?check_suite_focus=true and https://github.com/linux-system-roles/metrics/runs/7440737046?check_suite_focus=true you can see that the code somehow has knowledge of the main branch - it knows to use "main" or "master" somehow.  So my commit also includes some debugging - maybe we can use `$GITHUB_REF` or `$GITHUB_REF_NAME`

There's also the method that the checkout action uses - https://github.com/actions/checkout/blob/v2.4.2/src/git-command-manager.ts#L201 - where `repositoryUrl` comes from env `$GITHUB_SERVER_URL`/`$GITHUB_REPOSITORY`
see https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables for the complete list of env. vars. available to a github action

nhosoi (approved):
lgtm

richm:
> > [richm](https://github.com/richm) closed this [18 minutes ago](https://github.com/linux-system-roles/firewall/pull/98#event-7073656275)
> 
> May I ask why this pr was closed? Is it accidental?

It is being replaced by https://github.com/linux-system-roles/firewall/pull/99

PR #99, opened 2022-07-27
Use GITHUB_REF_NAME as name of push branch; fix error in branch detection [citest skip]

richm:
We need to get the name of the branch to which CHANGELOG.md was pushed.
For now, it looks as though `GITHUB_REF_NAME` is that name.  But don't
trust it - first, check that it is `main` or `master`.  If not, then use
a couple of other methods to determine what is the push branch.

### DIFF
--- a/.github/workflows/changelog_to_tag.yml
+++ b/.github/workflows/changelog_to_tag.yml
@@ -1,5 +1,5 @@
 # yamllint disable rule:line-length
-name: Pushing CHANGELOG.md triggers tagging
+name: Pushing CHANGELOG.md triggers tag, release, and Galaxy publish
 on:  # yamllint disable-line rule:truthy
   push:
     branches:
@@ -10,7 +10,7 @@ on:  # yamllint disable-line rule:truthy
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # This token is provided by Actions, you do not need to create your own token
 jobs:
-  tagging:
+  tag_release_publish:
     runs-on: ubuntu-latest
     steps:
       - name: checkout PR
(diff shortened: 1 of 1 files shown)


richm:
@nhosoi I ran this previously with storage - here is the log - https://github.com/linux-system-roles/storage/runs/7542318420?check_suite_focus=true

.github/workflows/changelog_to_tag.yml
```diff
@@ -19,30 +19,44 @@ jobs:
...
           done
-          # Get the main branch name, "master" or "main".
-          _branch=$( git branch -r | grep -o 'origin/HEAD -> origin/.*$' | \
-                     awk -F'/' '{print $3}' )
+          # Get name of the branch that the change was pushed to
+          _branch="${GITHUB_REF_NAME:-}"
+          if [ "$_branch" = master ] || [ "$_branch" = main ]; then
+              echo Using branch name ["$_branch"] as push branch
+          else
+              echo WARNING: GITHUB_REF_NAME ["$_branch"] is not main or master
```
nhosoi, line 48: I wonder if $GITHUB_REF_NAME is not `main` and `master`, it matches the one retrieved by the next line 49? Maybe, the line 49 is not needed? (I'd think issuing the warning at line 48 is a good idea.)

richm: > I wonder if $GITHUB_REF_NAME is not `main` and `master`, it matches the one retrieved by the next line 49?

I don't know.  My concern is that in the future `GITHUB_REF_NAME` may contain something that isn't a simple branch name e.g. maybe it would be `origin/main` or `refs/main`.  Or, maybe github decides to drop it altogether, in which case `$_branch` will be `""`.  In any of those cases, we would need to figure out the branch name some other way.

> Maybe, the line 49 is not needed? (I'd think issuing the warning at line 48 is a good idea.)

I hope line 49 is not needed.  I hope that we can trust that `GITHUB_REF_NAME` is always, now and in the future, the name of the push branch.

So another way to do this would be 
```
               echo ERROR: GITHUB_REF_NAME ["$_branch"] is not main or master
               exit 1
```
and that's that - either we use `GITHUB_REF_NAME` or exit with an error.  That would make the code much simpler.

nhosoi: I see. Indeed it's hard to predict if `GITHUB_REF_NAME` is guaranteed to be the same. I prefer your simpler proposal above, but the redundancy might pay off some day... Thanks for the explanation.

nhosoi (approved):
lgtm

PR #114, opened 2022-12-14
Add check for non-inclusive language

nhosoi:
Add a check for usage of terms and language that is considered non-inclusive. We are using the woke tool for this with a wordlist that can be found at https://github.com/linux-system-roles/tox-lsr/blob/main/src/tox_lsr/config_files/woke.yml

CHANGELOG.md - cleanup non-inclusive words.


richm (approved):

nhosoi:
@richm, thank you for your approval. If it looks ok, may I ask to merge this PR? Or we'd better have another approval? If so, whom shall we ask? Thanks.

PR #115, opened 2023-01-04
ansible-lint 6.x fixes

richm:
* All tasks must be named
* Task names must start with uppercase letter
* Use `true`/`false` instead of `yes`/`no`
* No spaces inside brackets and braces
* Add `changed_when: false` for command and shell in tests
* Documents must start with `---`
* Do not use yaml doc end `...`
* other changes


richm:
[citest]

richm:
[citest]

PR #120, opened 2023-01-20
cannot use distutils; use custom version

richm:
distutils is deprecated, and ansible-lint 2.14 throws an error about this.
There isn't really a suitable version parser replacement that works
without additional dependencies across python 2.7 - 3.11.  So, write
our own version parser.


richm:
[citest]

richm:
[citest]

Issue #138, opened 2023-04-18
Document that replacing previous rules causes a brief block of all new connections


myllynen:
When using `previous: replaced` there's a small time window when firewall is being reset and all new connections to the system are rejected. This should be at least documented if it is not feasible to avoid this as applying changes with this approach in production might cause temporary service failures during the operation. Thanks.

richm:
> When using `previous: replaced` there's a small time window when firewall is being reset and all new connections to the system are rejected. This should be at least documented if it is not feasible to avoid this as applying changes with this approach in production might cause temporary service failures during the operation. Thanks.

@myllynen how about this? https://github.com/linux-system-roles/firewall/pull/141

@vrindle @erig0 Is it feasible to avoid this?  How is using `previous: replaced` different than other firewalld tasks?

erig0:
> @erig0 Is it feasible to avoid this? How is using previous: replaced different than other firewalld tasks?

If the role is doing a `--reload`, then no.

During firewalld reload _existing_ connections continue to pass traffic (accepted), but while the rules are reloading _new_ connections are denied (rejected). I don't think there is any way to avoid this. You certainly don't want to blindly accept packets during a reload.

myllynen:
The fix looks good, closing.

PR #141, opened 2023-04-25
docs: Add note about using previous: replaced and temporary service failures

richm:
Add a note to the README about the use of `previous: replaced` and that it can
cause temporary service outages to the node being managed.
https://github.com/linux-system-roles/firewall/issues/138

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,6 +572,11 @@ useful if you have existing machines that may have existing firewall
 configuration, and you want to make all of the firewall configuration the same
 across all of the machines.
 
+*WARNING*: When using this option, there's a small time window when firewall is
+being reset and all new connections to the system are rejected.  Existing
+connections will be unaffected. Applying changes with this option in production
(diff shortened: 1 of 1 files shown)


myllynen (commented):
Thanks, looks good!

README.md
```diff
@@ -572,6 +572,11 @@ useful if you have existing machines that may have existing firewall
 configuration, and you want to make all of the firewall configuration the same
 across all of the machines.
 
+*WARNING*: When using this option, there's a small time window when firewall is
+being reset and all new connections to the system are rejected. Applying changes
+with this option in production might cause temporary service failures during the
+operation.
```
erig0, line 578: Might be worth adding that existing connections will still pass traffic.

richm: please see the new text

erig0 (approved):
lgtm. thanks!

PR #148, opened 2023-06-20
fix: Don't install python(3)-firewall it's a dependency of firewalld

spetrosi:
Enhancement: The role now does not run tasks to install python-firewall or python3-firewall based on installed python version.

Reason:  python-firewall or python3-firewall is pulled automatically by dnf and yum when installing firewalld.
The issue is that when I install python3 on EL 7, the role then fails with "No package matching 'python3-firewall' found available, installed or updated". It sees python3 present on the system and tries to install python3-firewall, which is not available on EL 7.

Result: The role doesn't fail on EL 7 when python3 is installed on the managed node.

### DIFF
--- a/tasks/firewalld.yml
+++ b/tasks/firewalld.yml
@@ -9,15 +9,3 @@
   package:
     name: "{{ __firewall_packages_base }}"
     state: present
-
-- name: Install python-firewall
-  package:
-    name: "{{ __firewall_packages_python2 }}"
-    state: present
-  when: ansible_python_version is version('3', '<')
-
-- name: Install python3-firewall
-  package:
-    name: "{{ __firewall_packages_python3 }}"
-    state: present
-  when: ansible_python_version is version('3', '>=')
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -8,7 +8,5 @@ __firewall_required_facts:
   - python_version
 
 __firewall_packages_base: [firewalld]
-__firewall_packages_python2: [python-firewall]
-__firewall_packages_python3: [python3-firewall]
 
 __firewall_service: firewalld


spetrosi:
[citest]

.github/workflows/commitlint.yml
```diff
@@ -25,6 +25,7 @@ jobs:
         run: npm install @commitlint/config-conventional @commitlint/cli
 
       - name: Run commitlint on PR title
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
         run: >-
-          echo '${{ github.event.pull_request.title }}' |
-          npx commitlint --verbose
+          echo "$PR_TITLE" npx commitlint --verbose
```
erig0, line 31: It looks like the pipe, `|`, was dropped.

erig0:
> there must have been some reason we explicitly installed the python firewall library

IMO it's the more correct dependency. Ansible is probably using the "client" classes in the python3-firewall module.

That being said, firewalld _always_ depends upon python3-firewall (or python-firewall). Most of the "daemon" code is actually in python3-firewall; alongside the client code.

.github/workflows/commitlint.yml
richm: good catch

richm:
> > there must have been some reason we explicitly installed the python firewall library
> 
> IMO it's the more correct dependency. Ansible is probably using the "client" classes in the python3-firewall module.

Whatever this is:
```python
    from firewall.client import (
        FirewallClient,
        Rich_Rule,
        FirewallClientZoneSettings,
        FirewallClientServiceSettings,
    )
```
and
```python
    from firewall.core.fw_nm import (
        nm_is_imported,
        nm_get_connection_of_interface,
        nm_get_zone_of_connection,
        nm_set_zone_of_connection,
        nm_get_interfaces,
        nm_get_client,
    )
```

> 
> That being said, firewalld _always_ depends upon python3-firewall (or python-firewall). Most of the "daemon" code is actually in python3-firewall; alongside the client code.
> 
> I can't come up with a reason why depending on `firewalld` instead of `python3-firewall` would cause a problem.

Ok - and it looks like the unit and integration tests would agree.   I think we can merge this (with the npx commitlint pipe fix).  Thanks!

erig0:
>  I think we can merge this (with the npx commitlint pipe fix).

Agree.

Issue #136, opened 2023-04-12
Disable nftables service


myllynen:
When using the firewall role only the firewalld service, not nftables, should be running to avoid any conflicts. While it's trivial to disable/mask nftables in a separate task it would be nice to have the role to do the same and thus avoid any risk for issues. Thanks.

BrennanPaciorek:
Recent PR added the variable `firewall_disable_conflicting_services` that can be set to check for and disable any conflicting services (currently disables iptables, nftables, and ufw). You will want to enable this on your first task running the role in each playbook. Thank you for reporting this issue.

Issue #103, opened 2022-09-02
when firewalld.service is masked, firewall role fails


jhutar:
**What happened**:
In one lab from where I'm taking systems from time to time they provide RHEL8 systems with firewalld.service masked by default. Running firewall role on them fails.

Maybe this is not a bug report, maybe it is not even a feature request, but still I though I report it as capability to deal with this might add another bit of resiliency.

```
# systemctl status firewalld
● firewalld.service
   Loaded: masked (Reason: Unit firewalld.service is masked.)
   Active: inactive (dead)
```

**What you expected to happen**:
Given firewall role decided to use firewalld as a provider, it should be able to handle this situation.

**How to reproduce it (as minimally and precisely as possible)**:
I think having this:
```
# rpm -qa | grep -e iptables -e firewalld | sort
firewalld-0.9.3-13.el8.noarch
firewalld-filesystem-0.9.3-13.el8.noarch
iptables-1.8.4-22.el8.x86_64
iptables-ebtables-1.8.4-22.el8.x86_64
iptables-libs-1.8.4-22.el8.x86_64
```
and firewalld masked and config like this:
```
    - role: linux-system-roles.firewall
      vars:
        firewall:
          - service: "RH-Satellite-6"
            state: enabled
```

**Anything else we need to know?**:
N/A

**Environment**:
```
# yum info firewalld
Updating Subscription Management repositories.
Satellite 6.12 RHEL8                            163 kB/s | 3.0 kB     00:00    
Satellite Maintenance 6.12 RHEL8                443 kB/s | 2.3 kB     00:00    
Installed Packages
Name         : firewalld
Version      : 0.9.3
Release      : 13.el8
Architecture : noarch
Size         : 2.0 M
Source       : firewalld-0.9.3-13.el8.src.rpm
Repository   : @System
From repo    : rhel-8-for-x86_64-baseos-rpms
Summary      : A firewall daemon with D-Bus interface providing a dynamic
             : firewall
URL          : http://www.firewalld.org
License      : GPLv2+
Description  : firewalld is a firewall service daemon that provides a dynamic
             : customizable firewall with a D-Bus interface.
```

```
# cat /etc/firewalld/firewalld.conf | grep FirewallBackend
# FirewallBackend
FirewallBackend=nftables
```

```
$ ansible --version
ansible [core 2.12.7]
  config file = /etc/ansible/ansible.cfg
[...]

PR #154, opened 2023-07-12
fix: unmask firewalld on run, disable conflicting services

BrennanPaciorek:
Enhancement:
Role will now always attempt to unmask on role execution

add variable 'firewall_disable_conflicting_services' to give the option of disabling of known conflicting services
- Set to false by default

Update README to document the following behavior of the system role:
- linux-system-roles.firewall will attempt to install, unmask, and enable firewalld
- linux-system-roles.firewall can attempt to disable directly conflicting services to firewalld
  - and that is enabled by setting the variable 'firewall_disable_conflicting_services' to true
  - list of conflicting services present in vars/main.yml
  
 test cases for these changes in tests/tests_default.yml

Reason:
role currently fails if firewalld was masked on run
conflicting services have the potential to cause errors on role run
- set to false by default due to runtime overhead associated with disabling conflicting services. An example of where this overhead may be a problem is our integration tests that have no need to use the feature.
- Reason for specific implementation - ansible.builtin.service module fails when run to manage services that are not installed on the system, causing errors. While ignoring errors is a potential solution, it seemed like an improper solution as it would not be able to differentiate between an installed service that failing to be stopped and disabled vs a disable that failed due to not being installed.

Result:
- role no longer fails if firewalld is masked
- users have the option to disable conflicting services (iptables.service, nftables.service, ufw.service respectively)

- Addresses GitHub Issues: #103, #136

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -12,26 +12,13 @@
     state: stopped
     enabled: false
   loop: "{{ __firewall_conflicting_services }}"
+  vars:
+    __service_name: "{{ (ansible_facts.service_mgr == 'systemd') |
+      ternary(item ~ '.service', item) }}"
   when:
     - firewall_disable_conflicting_services | bool
-    - (
-        ansible_facts.service_mgr == "systemd" and
-        item + ".service" | string in ansible_facts.services
-      )
-        or
-      (
-        ansible_facts.service_mgr != "systemd" and
-        item | string in ansible_facts.services
-      )
-    - (
-        ansible_facts.service_mgr == "systemd" and
-        ansible_facts.services[item + ".service"]["status"] == "enabled"
-      )
-        or
-      (
-        ansible_facts.service_mgr != "systemd" and
-        ansible_facts.services[item]["status"] == "enabled"
-      )
+    - __service_name | string in ansible_facts.services
+    - ansible_facts.services[__service_name]["status"] == "enabled"
 
 - name: Unmask firewalld service
   systemd:
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -2,6 +2,43 @@
 - name: Setup firewalld
   include_tasks: firewalld.yml
 
+- name: Collect service facts
+  service_facts:
+  when: firewall_disable_conflicting_services | bool
+
+- name: Attempt to stop and disable conflicting services
+  service:
+    name: "{{ item }}"
+    state: stopped
+    enabled: false
+  loop: "{{ __firewall_conflicting_services }}"
+  when:
+    - firewall_disable_conflicting_services | bool
+    - (
+        ansible_facts.service_mgr == "systemd" and
+        item + ".service" | string in ansible_facts.services
+      )
+        or
+      (
+        ansible_facts.service_mgr != "systemd" and
+        item | string in ansible_facts.services
+      )
+    - (
+        ansible_facts.service_mgr == "systemd" and
+        ansible_facts.services[item + ".service"]["status"] == "enabled"
+      )
+        or
+      (
+        ansible_facts.service_mgr != "systemd" and
+        ansible_facts.services[item]["status"] == "enabled"
+      )
+
+- name: Unmask firewalld service
+  systemd:
+    name: "{{ __firewall_service }}"
+    masked: false
+  when: ansible_facts.service_mgr == "systemd"
+
 - name: Enable and start firewalld service
   service:
     name: "{{ __firewall_service }}"
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -6,7 +6,13 @@ __firewall_firewalld_conf: "{{ __firewall_firewalld_dir }}/firewalld.conf"
 __firewall_required_facts:
   - python_interpreter
   - python_version
+  - service_mgr
 
 __firewall_packages_base: [firewalld]
 
 __firewall_service: firewalld
+
+__firewall_conflicting_services:
+  - nftables
+  - iptables
+  - ufw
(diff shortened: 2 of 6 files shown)


tasks/main.yml
```diff
@@ -2,6 +2,27 @@
 - name: Setup firewalld
   include_tasks: firewalld.yml
 
+- name: Collect Service Facts
+  service_facts:
+  when: firewall_disable_conflicting_services
```
richm, line 7:
```suggestion
  when: firewall_disable_conflicting_services | bool
```
see https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style#ansible-guidelines "Use the | bool filter when using bare variables"

tasks/main.yml
```diff
@@ -2,6 +2,27 @@
...
+
+- name: Attempt to stop and disable conflicting services
+  service:
+    name: "{{ item }}"
+    state: stopped
+    enabled: false
+  with_items: "{{ __conflicting_services }}"
+  when:
+    - firewall_disable_conflicting_services
```
richm, line 16:
```suggestion
    - firewall_disable_conflicting_services | bool
```

vars/main.yml
```diff
@@ -6,7 +6,13 @@ __firewall_firewalld_conf: "{{ __firewall_firewalld_dir }}/firewalld.conf"
 __firewall_required_facts:
   - python_interpreter
   - python_version
+  - service_mgr
 
 __firewall_packages_base: [firewalld]
 
 __firewall_service: firewalld
+
+__conflicting_services:
```
richm, line 15:
```suggestion
__firewall_conflicting_services:
```

tasks/main.yml
```diff
@@ -2,6 +2,27 @@
...
+- name: Collect Service Facts
+  service_facts:
+  when: firewall_disable_conflicting_services
+
+- name: Attempt to stop and disable conflicting services
+  service:
+    name: "{{ item }}"
+    state: stopped
+    enabled: false
+  with_items: "{{ __conflicting_services }}"
```
richm, line 14:
```suggestion
  loop: "{{ __firewall_conflicting_services }}"
```

tasks/main.yml
```diff
@@ -2,6 +2,27 @@
...
+  when:
+    - firewall_disable_conflicting_services
+    - item + ".service" | string in ansible_facts.services
+    - ansible_facts.services[item + ".service"]["status"] == "enabled"
```
richm, line 18: note that the use of `.service` implies a systemd system - I think this is ok for now

richm:
> This may have become too complex to be just a bug fix, but if we are disabling nftables, I figured it would make sense to disable all known conflicting services and make it an argument for the role to reduce overhead in the cases where doing so is not necessary.
> 
> I can additionally separate the two fixes into separate PRs if necessary.

I think it's fine to have a single pr

tasks/main.yml
BrennanPaciorek: I will change this condition to account for other init systems where the service is simply named {{ item }}

BrennanPaciorek: Will push changes resolving all of the above issues, my local integration tests are currently stalling on firewalld installation

richm: > Will push changes resolving all of the above issues, my local integration tests are currently stalling on firewalld installation

What platform?

BrennanPaciorek: Reverting my changes to when tests did not stall also did not work. Maybe its an external issue?

BrennanPaciorek: Verbatim command I used for testing was `tox -e qemu-ansible-test-2.14 -- --image-name centos-8 tests/tests_default.yml`
Same issue happened when I ran it with the centos-9 image.

I think the mirror the role was trying to use was just being particularly difficult.

tasks/main.yml
```diff
@@ -2,6 +2,43 @@
...
+      )
+    - (
+        ansible_facts.service_mgr == "systemd" and
+        ansible_facts.services[item + ".service"]["status"] == "enabled"
+      )
+        or
+      (
+        ansible_facts.service_mgr != "systemd" and
+        ansible_facts.services[item]["status"] == "enabled"
+      )
```
BrennanPaciorek, line 34: Changed the conditions for disabling conflicting processes to specifically check the service manager, and only append ".service" if running on systemd.  Does this seem like a sufficient implementation for the check @richm?

richm: I think it would be cleaner by using an intermediate var, and I don't think checking for state is necessary:
```
- name: Attempt to stop and disable conflicting services
  service:
    name: "{{ item }}"
    state: stopped
    enabled: false
  loop: "{{ __firewall_conflicting_services }}"
  vars:
    __service_name: "{{ (ansible_facts.service_mgr == "systemd") |
      ternary(item ~ '.service', item) }}"
  when:
    - firewall_disable_conflicting_services | bool
```
Isn't the `service` module idempotent?  Or is it a purely performance issue to check explicitly first?

BrennanPaciorek: service module fails if the service is not installed when disabling is attempted in either check or diff mode.

BrennanPaciorek: Recent push implemented intermediate var for the service name.

richm:
[citest]

richm:
We need to move the test from tests_default.yml - that test is only intended to run the role with no parameters.  If you can alter another test, please do, otherwise, we will require a new test file.

BrennanPaciorek:
> We need to move the test from tests_default.yml - that test is only intended to run the role with no parameters. If you can alter another test, please do, otherwise, we will require a new test file.

Done. (new test file created)

richm:
[citest]

richm (approved):

Issue #140, opened 2023-04-25
Provide example or method to show differences between defined vs in-use configuration


myllynen:
Since replacing previous rules causes firewalld restart it is not suitable for most production environments. In case not replacing previous rules the defined configuration may not be what is in use due to local changes or previous playbook runs with different configuration. An organization might be adhering to security standards that require listing or verifying that only certain ports and services are open in firewall.

There should be a way to display differences between currently defined configuation vs currently in-use configuration. Preferably this would be a configurable step during firewall configuration which would show differences as part of a playbook run after configuring firewall. This should also be doable in check mode. Alternatively, a separate playbook or at least a document example should be available to allow organizations avoid reinventing the wheel here.

Thanks.

richm:
> Since replacing previous rules causes firewalld restart it is not suitable for most production environments. In case not replacing previous rules the defined configuration may not be what is in use due to local changes or previous playbook runs with different configuration. An organization might be adhering to security standards that require listing or verifying that only certain ports and services are open in firewall.
> 
> There should be a way to display differences between currently defined configuation vs currently in-use configuration. Preferably this would be a configurable step during firewall configuration which would show differences as part of a playbook run after configuring firewall. This should also be doable in check mode. Alternatively, a separate playbook or at least a document example should be available to allow organizations avoid reinventing the wheel here.

That's what I thought https://github.com/linux-system-roles/firewall#available-ansible-facts was supposed to be used for.  Is that not suitable for your purposes?

> 
> Thanks.

myllynen:
Those facts are definitely helpful, however here I'd consider them more as a necessary building block than a complete solution.

An example could be a case where a large number of systems have two zones configured and few rich rules in use. Having readily available Ansible to print out the differences for those servers with differences a readable manner would be very helpful. For instance, if someone opened the HTTP service in addition to the currently configured and expected HTTPS service in the public zone and created an extra rich rule in another zone, how to "easily" and in compact manner provide that information? Thanks.

richm:
It seems to me that this is what check mode is intended for.  Does the firewall role check mode not work for this situation?  Or is it that the output of check mode is not easily consumable, and that you would like the data formatted in a more "user friendly" format?

myllynen:
Sorry for the delay with this one, I somehow missed your questions earlier.

Wrt check mode, I noticed that it doesn't work when replacing previous rules, filed #151 about that.

The check mode works very well when adding something. For instance, consider this simple example:

```
    firewall:
[...]

PR #159, opened 2023-07-18
fix: reload on resetting to defaults

BrennanPaciorek:
Enhancement:
Make resetting to defaults reload instead of restart firewalld

Reason:
Reloading in firewalld should successfully complete the configuration reset, and restarting adds downtime which can be used to open a connection that persists after firewalld has finishes restarting; this connection can be used to bypass firewall rules, since firewalld will not block traffic from active connections.

Result:
Minimal downtime when using `previous: replaced`

Addresses an issue brought up in #140, where due to the restart on resetting to defaults, the feature may not be suitable for production environments.


richm (approved):

BrennanPaciorek:
> * If `nft list ruleset` output is less predictable than I think, testing it may be more difficult than this.

It appears that restart and reload produce functionally identical configurations, but the rules tend to be reordered slightly within the same rule chains, so I'll need to find a more accurate way to compare two configurations.

richm:
> > how can we test this?
> 
> If you mean verifying that this will not introduce bugs into the feature, we could use `firewall-cmd` with ansible.builtin.command. This would involve:
> 
>     * Using the current implementation of `previous: replaced` to reset the firewall configuration
> 
>     * getting some representation of the active firewall implementation (something like a hash of `nft list ruleset`)
> 
>     * restarting firewalld service instead of reloading
> 
>     * get the representation of the active firewall implementation again
> 
>     * Compare the two stored values
>       If `nft list ruleset` output is less predictable than I think, testing it may be more difficult than this.
> 
> 
> If you mean testing minimal downtime, `systemctl reload firewalld` uses firewalld's reload implementation to reload configuration from files and `systemctl restart firewalld` turns the firewall off and back on (leaving the firewall off for a few seconds, which is usually a bad idea for a production environment).

I mean the latter - check that the fix ensures minimal downtime - is there a way we can automate this - if not, how do we tell QE what to test?

BrennanPaciorek:
> is there a way we can automate this

Testing service downtime (or uptime) wouldn't work, because with `reload`, the service technically does not stop.

An idea would to be change a zone to DROP (or set a ICMP block) and place an interface on this zone, then spam a set number of packets that would be dropped under this do. If the packets get sent back, it can be taken to mean the firewall is down, if they are lost, you can take it to mean the firewall is up and dropping the packets. Whichever loses the most packets can be understood as having the least downtime. Does this seem like a valid test?
-

BrennanPaciorek:
> They both seem like valid tests - by "simple http server", maybe `nc -l -p 80 -o file`? then we could look at `file` and see if packets were dropped?

`nc` would be better for mimicking production environments, while `ping` is probably a more convenient tool for getting the necessary metrics (%packet loss, estimated downtime)

One issue so far in implementing this in a tox-friendly manner that arises is that firewalld accepts outbound traffic on the loopback interface before evaluating policies (where outbound traffic is blocked), and disabling this rule takes some time (~100ms), which may result in more "downtime" than a production environment, where this limitation of needing to use the loopback interface for tests does not exist.

The other issue with these tests is that using `previous: replaced` feature would undo the firewall rules needed to test the effectiveness of the modified feature. would simple comparisons in downtime between `systemctl reload firewalld.service` and `systemctl restart firewalld` suffice for testing this?

richm:
> > They both seem like valid tests - by "simple http server", maybe `nc -l -p 80 -o file`? then we could look at `file` and see if packets were dropped?
> 
> `nc` would be better for mimicking production environments, while `ping` is probably a more convenient tool for getting the necessary metrics (%packet loss, estimated downtime)
> 
> One issue so far in implementing this in a tox-friendly manner is that firewalld accepts outbound traffic on the loopback interface before evaluating policies (where outbound traffic is blocked), and disabling this rule takes some time (~100ms), which may result in more "downtime" than a production environment (or even a test environment with less restrictions), where this limitation of needing to use the loopback interface for tests does not exist.
> 
> The other issue with these tests is that using `previous: replaced` feature would undo the firewall rules needed to test the effectiveness of the modified feature. Would comparisons in downtime between `systemctl reload firewalld.service` and `systemctl restart firewalld.service` suffice for testing this?

Yes, I think that would be fine.

> If not, we can use a modified reset script to re-add any required rules before restarting or reloading firewalld.

tests/files/test_ping.sh
```diff
@@ -0,0 +1,51 @@
...
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping2 2>/dev/null &
+pid=$!
+trap "rm -f /tmp/ping2" EXIT
+podman exec test-firewalld systemctl reload firewalld.service
+wait $pid
+
+# Print Results
+cat /tmp/ping0 | tail -2 | head -1 | awk '{print $4}'
+cat /tmp/ping1 | tail -2 | head -1 | awk '{print $4}'
+cat /tmp/ping2 | tail -2 | head -1 | awk '{print $4}'
```
BrennanPaciorek: tail -2 <file>

tests/files/test_ping.sh
```diff
@@ -0,0 +1,51 @@
...
+EOF
+
+trap "rm -f /tmp/Containerfile" EXIT
+
+# Initial container setup #
+{
+  podman network create --subnet 172.16.1.0/24 --gateway 172.16.1.1 --interface-name podmanbr0 podmanbr0
+  trap "podman network rm podmanbr0" EXIT
+  imageid=$(podman build -q /tmp)
+  podman run -d --privileged --net podmanbr0 --ip 172.16.1.2 --name test-firewalld --rm $imageid /usr/lib/systemd/systemd || exit 1
```
BrennanPaciorek: "$imageid"

BrennanPaciorek:
Okay I've implemented the described test in a way that is compatible with the integration tests, I can move this to another file, but the test takes some time and probably shouldn't be run every PR, since its not testing any code prone to change.

tests/files/test_ping.sh
```diff
@@ -0,0 +1,51 @@
...
+EOF
+
+trap "rm -f /tmp/Containerfile" EXIT
+
+# Initial container setup #
+{
+  podman network create --subnet 172.16.1.0/24 --gateway 172.16.1.1 --interface-name podmanbr0 podmanbr0
+  trap "podman network rm podmanbr0" EXIT
+  imageid=$(podman build -q /tmp)
+  podman run -d --privileged --net podmanbr0 --ip 172.16.1.2 --name test-firewalld --rm "$imageid" /usr/lib/systemd/systemd || exit 1
```
richm, line 21:
```suggestion
  podman run -d --rm --privileged --net podmanbr0 --ip 172.16.1.2 --name test-firewalld --rm "$imageid" /usr/lib/systemd/systemd || exit 1
```

tests/files/test_ping.sh
```diff
@@ -0,0 +1,51 @@
...
+  # firewall-cmd reload waits for dbus response, systemctl will not
+  podman exec test-firewalld firewall-cmd --reload
+} > /dev/null 2>/dev/null
+# The following ping should have 100% packet loss
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping0 2>/dev/null
+trap "rm -f /tmp/ping0" EXIT
+
+# Begin downtime comparision #
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping1 2>/dev/null &
+pid=$!
```
richm, line 36:
```suggestion
pid="$!"
```
make shellcheck happy

tests/files/test_ping.sh
```diff
@@ -0,0 +1,51 @@
...
+trap "rm -f /tmp/ping1" EXIT
+podman exec test-firewalld systemctl restart firewalld.service
+wait $pid
```
richm, line 39:
```suggestion
wait "$pid"
```

tests/files/test_ping.sh
```diff
@@ -0,0 +1,51 @@
...
+
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping2 2>/dev/null &
+pid=$!
```
richm, line 42:
```suggestion
pid="$!"
```

tests/files/test_ping.sh
```diff
@@ -0,0 +1,51 @@
...
+trap "rm -f /tmp/ping2" EXIT
+podman exec test-firewalld systemctl reload firewalld.service
+wait $pid
```
richm, line 45:
```suggestion
wait "$pid"
```

.sanity-ansible-ignore-2.12.txt
```diff
@@ -2,3 +2,4 @@ plugins/modules/firewall_lib.py validate-modules:missing-gplv3-license
 plugins/modules/firewall_lib_facts.py validate-modules:missing-gplv3-license
 plugins/modules/firewall_lib.py validate-modules:missing-examples
 roles/firewall/files/get_files_checksums.sh shebang!skip
+files/test_ping.sh shebang!skip
```
richm, line 5: `tests/firewall/files/test_ping.sh`

ansible-test is only run on the converted collection - in the collection the path is `tests/firewall/files/test_ping.sh`

tests/files/test_ping.sh
BrennanPaciorek: The --rn option was already provided, should I move it?

richm: sorry, I missed that

richm:
[citest]

tests/files/test_ping.sh
```diff
@@ -0,0 +1,51 @@
...
+pid="$!"
+trap "rm -f /tmp/ping1" EXIT
+podman exec test-firewalld systemctl restart firewalld.service
+wait "$pid"
+
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping2 2>/dev/null &
+pid="$!"
+trap "rm -f /tmp/ping2" EXIT
+podman exec test-firewalld systemctl reload firewalld.service
+wait "$pid"
```
BrennanPaciorek, line 45: The order of this sometimes introduces a fail case where a connection is established from `systemctl restart firewalld.service`, I swapped the two tests locally, but am waiting for CI tests to run.

BrennanPaciorek:
Locally, I can't seem to reproduce the same error that is showing on these CI tests. Maybe it has to do with the order in which the tests are being run?

richm:
> Locally, I can't seem to reproduce the same error that is showing on these CI tests. Maybe it has to do with the order in which the tests are being run?

Yes, that must be what it is

tests/files/test_ping.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+# Author - Brennan Paciorek <bpaciore@redhat.com>
+# Description - Benchmark firewalld downtime while reloading and while restarting
+# by measuring how many packets are dropped while firewalld is restarting/reloading
+
```
richm, line 5: `set -euo pipefail`

tests/files/test_ping.sh
```diff
@@ -0,0 +1,52 @@
...
+EXPOSE 31337
+CMD /usr/lib/systemd/systemd 
+EOF
+
+trap "rm -f /tmp/Containerfile" EXIT
+
+# Initial container setup #
+{
+  podman network create --subnet 172.16.1.0/24 --gateway 172.16.1.1 --interface-name podmanbr0 podmanbr0
+  trap "podman network rm podmanbr0" EXIT
```
richm, line 19: hmm - I don't think `trap` is additive - I think this `trap` will erase the previous `trap`:
```
bash -c 'trap "echo trap1" EXIT; trap "echo trap2" EXIT'
trap2
```
If you want to have multiple commands in a `trap`, define a `cleanup` shell function which does all of the cleanup:
```bash
cleanup() {
  rm -f /tmp/Containerfile
  podman network rm podmanbr0 || :
  podman stop test-firewalld || :
  rm -f /tmp/ping0
  rm -f /tmp/ping2
  rm -f /tmp/ping1
}
...
trap "cleanup" EXIT
```

BrennanPaciorek: You are right, changing this now, just have to debug the cleanup function, since it ignores the `|| :`, most likely due to pipefail

tests/files/test_ping.sh
```diff
@@ -0,0 +1,52 @@
...
+wait "$pid"
+
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping2 2>/dev/null &
+pid="$!"
+trap "rm -f /tmp/ping1" EXIT
+podman exec test-firewalld systemctl restart firewalld.service
+wait "$pid"
+
+# Print Results
+tail -2 /tmp/ping0 | head -1 | awk '{print $4}'
```
richm, line 49: Are these causing the pipefail?  You might be able to just do
`tail -2 /tmp/ping0 | head -1 | awk '{print $4}' || :`
if that doesn't work, then you'll have to disable pipefail here:
```bash
set +o pipefail
tail -2 /tmp/ping0 | head -1 | awk '{print $4}'
...
```

BrennanPaciorek: Maybe, I'm thinking its the ping commands (which return one if the command receives less than they send I think)

tests/files/test_ping.sh
```diff
@@ -0,0 +1,52 @@
...
+  sleep 5 # Wait reasonable amount of time for container to start services
+  
+  # Firewall rule setup #
+  podman exec test-firewalld firewall-cmd --permanent --add-icmp-block "echo-request"
+  # firewall-cmd reload waits for dbus response, systemctl will not
+  podman exec test-firewalld firewall-cmd --reload
+} > /dev/null 2>/dev/null
+
+# The following ping should have 100% packet loss
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping0 2>/dev/null
```
richm, line 32:
```suggestion
ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping0 2>/dev/null || :
```

tests/files/test_ping.sh
```diff
@@ -0,0 +1,52 @@
...
+trap "rm -f /tmp/ping0" EXIT
+
+# Begin downtime comparision #
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping1 2>/dev/null &
```
richm, line 36:
```suggestion
ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping1 2>/dev/null || : &
```

tests/files/test_ping.sh
```diff
@@ -0,0 +1,52 @@
...
+pid="$!"
+trap "rm -f /tmp/ping2" EXIT
+podman exec test-firewalld systemctl reload firewalld.service
+wait "$pid"
+
+ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping2 2>/dev/null &
```
richm, line 42:
```suggestion
ping -c 500 -i 0.01 172.16.1.2 1>/tmp/ping2 2>/dev/null || : &
```

tests/files/test_ping.sh
richm: > Maybe, I'm thinking its the ping commands (which return one if the command receives less than they send I think)

Yes, I think that's it - see my suggestions above

tests/files/test_ping.sh
BrennanPaciorek: The script seems to really not care if `|| :` is there or not. It seems to just ignore the second side of || entirely. testing with `|| true` but they should be the same thing

richm: hmm - try running with `set -euxo pipefail` - the `-x` for trace might tell you where the problem is

I used the suggested code in my local test and it worked

BrennanPaciorek: Update, was just missing a `|| :`

richm:
[citest]

tests/tests_reload_on_reset.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Setup a vm with podman on it
+  hosts: all
```
richm, line 3: skip test on el7
```yaml
  pre_tasks:
    - name: Test not supported on EL7
      meta: end_host
      when:
        - ansible_distribution in ['RedHat', 'CentOS']
        - ansible_distribution_major_version | int < 8
```

BrennanPaciorek: Done

BrennanPaciorek:
Modified test one more time, as some platforms running the test incorrectly reported packet loss due to the timeout I had set being too short.

richm (approved):

Issue #127, opened 2023-02-13
 /etc/firewalld/firewalld.conf is resetted, and all default comment are deleted


bdouxx:
**What happened**:

 /etc/firewalld/firewalld.conf is resetted, and all default comment are deleted

**What you expected to happen**:
 only parameters are modified 

**How to reproduce it (as minimally and precisely as possible)**:

```
---

- name: "[@@ firewalld @@]"
  ansible.builtin.include_role:
    name: linux-system-roles.firewall
  vars:
    firewall:
      - previous: replaced

      - zone: myownzone
        state: present

      - zone: myownzone
        service: [ssh]
        icmp_block_inversion: true
        icmp_block: [ echo-reply, echo-request ]
        rich_rule: rule family="ipv4" source address="192.168.1.95" service name="zabbix-agent" accept
        state: enabled
        target: "%%REJECT%%"
        permanent: true

      - set_default_zone: myownzone
        permanent: true
```

```
# cat /etc/firewalld/firewalld.conf

DefaultZone=myownzone
CleanupOnExit=yes
CleanupModulesOnExit=yes
Lockdown=no
IPv6_rpfilter=yes
IndividualCalls=no
LogDenied=off
FirewallBackend=nftables
FlushAllOnReload=yes
RFC3964_IPv4=yes
AllowZoneDrifting=yes
```

```
# cat /etc/firewalld/firewalld.conf.rpmnew
# firewalld config file

# default zone
# The default zone used if an empty zone string is used.
# Default: public
DefaultZone=public

# Clean up on exit
# If set to no or false the firewall configuration will not get cleaned up
# on exit or stop of firewalld.
# Default: yes
CleanupOnExit=yes

# Clean up kernel modules on exit
# If set to yes or true the firewall related kernel modules will be
# unloaded on exit or stop of firewalld. This might attempt to unload
# modules not originally loaded by firewalld.
# Default: yes
CleanupModulesOnExit=yes
....
```

why don't you use something like this:
```
- name: "[firewalld] AllowZoneDrifting=no"
  ansible.builtin.lineinfile:
    path: "/etc/firewalld/firewalld.conf"
    regexp: '^(AllowZoneDrifting=)(.*)'
    line: 'AllowZoneDrifting=no'
    insertafter: '#AllowZoneDrifting='
  notify:
    - reload firewalld
```

in the current ansible playbook, how to force AllowZoneDrifting=no?

richm:
Is it a problem that the comments are removed?

[...]

PR #162, opened 2023-07-21
feat: add new arg firewalld_conf, subarg allow_zone_drifting

BrennanPaciorek:
Enhancement:
firewall varable option firewalld_conf added to support modifying supported firewalld_conf arguments
- Sub-argument allow_zone_drifting modifies AllowZoneDrifting if possible via dbus
  - This should be supported in EL7+ due to backporting
- Associated Documentation added
- Unit tests for feature added to tests/unit/test_firewall_lib.py
- Integration test for feature in tests/tests_firewalld_conf.yml

Reason:
Many EL releases for firewalld set `AllowZoneDrifting=yes` as the default. This feature allows those who do not want this insecure feature set to disable the feature using the role.

Result:
AllowZoneDrifting can be changed to whatever the role user wants to change it to, provided it is not deprecated.

Will fail if the system is using a version of firewalld where `AllowZoneDrifting` has not been implemented. This should not be an issue on EL7 since the feature seems to have been backported to those systems, and should not be an issue on systems where the feature is removed because the dbus method for modifying the directive is still present (although it does not do anything).

- Addresses GitHub Issue #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ if support for their modification has been implemented.
 ```yaml
 firewall:
   - firewalld_conf:
-      allow_zone_drifting: no
+      allow_zone_drifting: false
     permanent: true
 ```
 
@@ -249,7 +249,7 @@ and no longer exists.
 ```yaml
 firewall:
   firewalld_conf:
-    allow_zone_drifting: yes
+    allow_zone_drifting: true
   permanent: true
 ```
 
--- a/README.md
+++ b/README.md
@@ -225,6 +225,30 @@ add it locally to `vars/main.yml`.
 
 The firewall role uses the variable `firewall` to specify the parameters.  This variable is a `list` of `dict` values.  Each `dict` value is comprised of one or more keys listed below. These are the variables that can be passed to the role:
 
+### firewalld_conf
+`firewalld_conf` can be used to modify directives in firewalld's configuration file (`/etc/firewalld/conf` by default)
+if support for their modification has been implemented.
+
+**`permanent: true` must always be set to run this option without error**
+
+```yaml
+firewall:
+  - firewalld_conf:
+      allow_zone_drifting: "no"
+    permanent: true
+```
+
+#### Supported Directives
+##### allow_zone_drifting
+
+Changes the AllowZoneDrifting directive, if the directive exists.
+
+Accepted values:
+- "yes"
+- "no"
+
+Quotations around "yes" and "no" are necessary, because ansible attempts to convert the argument into boolean values otherwise.
+
 ### set_default_zone
 
 The default zone is the zone that is used for everything that is not explicitly
--- a/README.md
+++ b/README.md
@@ -234,20 +234,24 @@ if support for their modification has been implemented.
 ```yaml
 firewall:
   - firewalld_conf:
-      allow_zone_drifting: "no"
+      allow_zone_drifting: no
     permanent: true
 ```
 
 #### Supported Directives
 ##### allow_zone_drifting
 
-Changes the AllowZoneDrifting directive, if the directive exists.
+Changes the AllowZoneDrifting directive.
 
-Accepted values:
-- "yes"
-- "no"
+This parameter will do nothing if AllowZoneDrifting has been deprecated
+and no longer exists.
 
-Quotations around "yes" and "no" are necessary, because ansible attempts to convert the argument into boolean values otherwise.
+```yaml
+firewall:
+  firewalld_conf:
+    allow_zone_drifting: yes
+  permanent: true
+```
 
 ### set_default_zone
 
--- a/library/firewall_lib.py
+++ b/library/firewall_lib.py
@@ -541,6 +541,23 @@ def parse_forward_port(module, item):
     return (_port, _protocol, _to_port, _to_addr)
 
 
+def parse_firewalld_conf(am, firewalld_conf):
+    allow_zone_drifting_removed = False
+    if firewalld_conf["allow_zone_drifting"] is not None:
+        if lsr_parse_version(FW_VERSION) >= lsr_parse_version("1.0.0"):
+            am.warn(
+                "AllowZoneDrifting is deprecated in this version of firewalld, unsetting parameter"
+            )
+            del firewalld_conf["allow_zone_drifting"]
+            allow_zone_drifting_removed = True
+        else:
+            if firewalld_conf["allow_zone_drifting"]:
+                firewalld_conf["allow_zone_drifting"] = "yes"
+            else:
+                firewalld_conf["allow_zone_drifting"] = "no"
+    return allow_zone_drifting_removed
+
+
 def set_the_default_zone(fw, set_default_zone):
     fw.setDefaultZone(set_default_zone)
 
@@ -629,11 +646,9 @@ def main():
     # Argument parse
     firewalld_conf = module.params["firewalld_conf"]
     if firewalld_conf:
-        if firewalld_conf["allow_zone_drifting"] is not None:
-            if firewalld_conf["allow_zone_drifting"]:
-                firewalld_conf["allow_zone_drifting"] = "yes"
-            else:
-                firewalld_conf["allow_zone_drifting"] = "no"
+        allow_zone_drifting_removed = parse_firewalld_conf(module, firewalld_conf)
+    else:
+        allow_zone_drifting_removed = False
     service = module.params["service"]
     short = module.params["short"]
     description = module.params["description"]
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -225,6 +225,30 @@ add it locally to `vars/main.yml`.
...
+      allow_zone_drifting: "no"
+    permanent: true
+```
+
+#### Supported Directives
+##### allow_zone_drifting
+
+Changes the AllowZoneDrifting directive, if the directive exists.
+
+Accepted values:
```
spetrosi, line 246: That's bad, vice versa these must be booleans, and you need to convert to the values you want in the code.

README.md
```diff
@@ -234,20 +234,24 @@ if support for their modification has been implemented.
 ```yaml
 firewall:
   - firewalld_conf:
-      allow_zone_drifting: "no"
+      allow_zone_drifting: no
```
richm, line 237:
```suggestion
      allow_zone_drifting: false
```
https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style#yaml-and-jinja2-syntax "Use true and false for boolean values in playbooks"

tests/tests_firewalld_conf.yml
```diff
@@ -3,56 +3,53 @@
...
-    - name: Start with clean configuration
+    - name: Try to disable zone drifting
       include_role:
         name: linux-system-roles.firewall
       vars:
         firewall:
           previous: replaced
           firewalld_conf:
-            allow_zone_drifting: "yes"
+            allow_zone_drifting: no
```
richm, line 30:
```suggestion
            allow_zone_drifting: false
```

tests/tests_firewalld_conf.yml
```diff
@@ -3,56 +3,53 @@
...
-      vars:
-        expected_result: >-
-          "{{ allow_zone_drifting_deprecated | ternary('no', 'yes') | quote }}"
+    - name: Check if AllowZoneDrifting is disabled if possible
       shell:
         cmd: >-
           set -o pipefail;
-          (grep AllowZoneDrifting /etc/firewalld/firewalld.conf
-          || echo "AllowZoneDrifting=no") | tail -1
+          grep AllowZoneDrifting /etc/firewalld/firewalld.conf | tail -1
```
richm, line 37: Do you want to check if `AllowZoneDrifting` is in the file but commented out?  e.g.
```
# AllowZoneDrifting=no
```
?

BrennanPaciorek: This is changing, but I want to get the actual directive for AllowZoneDrifting.

tests/tests_firewalld_conf.yml
```diff
@@ -53,3 +52,36 @@
...
+      register: result
+      changed_when: false
+      failed_when: result.rc == 2
+
+    - name: Fail if AllowZoneDrifting disabled if not deprecated
+      fail:
+        msg: "AllowZoneDrifting is disabled when it should be enabled"
+      when:
+        - allow_zone_drifting_not_deprecated
```
richm, line 78:
```suggestion
        - allow_zone_drifting_not_deprecated | bool
```

tests/tests_firewalld_conf.yml
```diff
@@ -53,3 +52,36 @@
...
+        - result.rc == 0
+        - result.stdout != "AllowZoneDrifting=yes"
+
+    - name: Fail if AllowZoneDrifting not found and not deprecated
+      fail:
+        msg: "AllowZoneDrifting directive not found in firewalld.conf"
+      when:
+        - allow_zone_drifting_not_deprecated
```
richm, line 86:
```suggestion
        - allow_zone_drifting_not_deprecated | bool
```

README.md
BrennanPaciorek: Makes sense, changed the parameter to accept booleans instead of string choices.

tests/tests_firewalld_conf.yml
BrennanPaciorek: Changing this to use `grep -Fx AllowZoneDrifting=yes /etc/firewalld/firewalld.conf` instead.

library/firewall_lib.py
```diff
@@ -914,6 +947,19 @@ def exception_handler(exception_message):
...
+        fw_config = fw.config()
+        if firewalld_conf.get("allow_zone_drifting") != fw_config.get_property(
+            "AllowZoneDrifting"
+        ):
+            if not module.check_mode:
+                fw_config.set_property(
+                    "AllowZoneDrifting", firewalld_conf.get("allow_zone_drifting")
+                )
+            changed = True
+            need_reload = True
```
BrennanPaciorek, line 961: `fw_config.get_property("AllowZoneDrifting")` will always return "no" on versions where AllowZoneDrifting is deprecated, and setting will do nothing in this case. This causes an idempotency issue when AllowZoneDrifting is deprecated and `firewall_conf.get("allow_zone_drifting") == "yes"` by setting `changed=true` when nothing has changed, as well as prompting a reload when none is needed.

Planned fix:
Check if firewalld version >= 1.0.0, then the module will remove firewalld_conf["allow_zone_drifting"], warning the user. If this causes an error at time of arg parse (ie not enough args supplied), the module will exit early.

Does this sound like a sufficient fix?

richm: > Does this sound like an ideal fix?

Yes, sounds good.

library/firewall_lib.py
```diff
@@ -541,6 +541,23 @@ def parse_forward_port(module, item):
     return (_port, _protocol, _to_port, _to_addr)
 
 
+def parse_firewalld_conf(am, firewalld_conf):
```
richm, line 544:
```suggestion
def parse_firewalld_conf(am, firewalld_conf):
```
suggest a more descriptive name like `check_allow_zone_drifting` or something similar

BrennanPaciorek: I can make a `check_allow_zone_drifting` function to be used within `parse_firewalld_conf`, but  I prefer to keep `parse_firewalld_conf` to check other arguments for `firewalld_conf` if they are added (ex. FirewallBackend).

richm: ok - it doesn't look like `parse_firewalld_conf` is actually parsing firewalld.conf - so I'm not sure what is the intention of `parse_firewalld_conf` or how to make it more generalized for other parameters

BrennanPaciorek: Does the most recent commit solve the issue?

BrennanPaciorek: > ok - it doesn't look like `parse_firewalld_conf` is actually parsing firewalld.conf - so I'm not sure what is the intention of `parse_firewalld_conf` or how to make it more generalized for other parameters

I see what you mean, this function checks the parameter `firewalld_conf`, should I change the name of the parameter to something like just `conf` so it is not conflated with the configuration file?

Then "parse" was probably a poor choice of words, so I changed the function name as well.

library/firewall_lib.py
```diff
@@ -693,14 +739,19 @@ def main():
...
             )
         )
     ):
-        module.fail_json(
-            msg="One of service, port, source_port, forward_port, "
-            "masquerade, rich_rule, source, interface, icmp_block, "
-            "icmp_block_inversion, target, zone or set_default_zone needs to be set"
-        )
+        if firewalld_conf_emptied:
+            return module.exit_json(changed=False, __firewall_changed=False)
```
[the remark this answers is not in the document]
BrennanPaciorek: It is the case where allow_zone_drifting is set on a version where it is unsupported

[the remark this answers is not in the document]
richm: I don't understand - `firewalld_conf_emptied` will only be true if `firewalld_conf` is set, but this code block cannot be reached if `firewalld_conf` is set?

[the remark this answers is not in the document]
BrennanPaciorek: It can be, I have a unit test explicitly testing the ability for this block to be reached.
- User sets firewalld_conf with `allow_zone_drifting`
- on `check_firewalld_conf`, `allow_zone_drifting` is removed because it would cause idempotency issues otherwise on firewalld versions where `AllowZoneDrifting` no longer exists
- firewalld_conf_emptied == True, because firewalld_conf is now a length 0 dictionary, making it also falsy
- Upon reaching the conditional, firewalld_conf is both falsy and firewalld_conf_emptied == true, so if no other parameters are set, module.exit_json() will execute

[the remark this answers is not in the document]
BrennanPaciorek: Though either way, if it needed explanation, I probably need a better way to do this.

library/firewall_lib.py
```diff
@@ -914,6 +964,19 @@
     changed = False
     need_reload = False
 
+    # firewalld.conf
+    if firewalld_conf:
+        fw_config = fw.config()
+        if not allow_zone_drifting_deprecated and firewalld_conf.get(
```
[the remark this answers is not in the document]
BrennanPaciorek: `allow_zone_drifting_deprecated` is guarenteed to be initialized, can/should this be supressed?

richm:
[citest]

richm:
[citest]

Issue #151, opened 2023-06-26
Check mode fails with replacing previous rules


myllynen:
When using `previous: replaced` and check mode, the playbook fails with the following error:

```
TASK [rhel-system-roles.firewall : Calculate what has changed] ****************
Monday 26 June 2023  12:10:46 +0300 (0:00:00.104)       0:00:12.707 ***********                                      
task path: /usr/share/ansible/roles/rhel-system-roles.firewall/tasks/main.yml:130
fatal: [192.168.122.170]: FAILED! =>                                                                                   
    msg: |-                                                                                                            
        The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'stdout_lines'. 'dict object' has no attribute 'stdout_lines'

        The error appears to be in '/usr/share/ansible/roles/rhel-system-roles.firewall/tasks/main.yml': line 130, column 7, but may
        be elsewhere in the file depending on the exact syntax problem.

        The offending line appears to be:


            - name: Calculate what has changed
              ^ here                                       
```

BrennanPaciorek:
Hello myllynen,

I am having difficulty replicating this error.

All of my testing of `previous: replaced` in check mode has failed to reproduce this (or any) error. I have attempted to purge config in check mode both with a default firewall configuration and with the following configuration active:
```yaml
- zone: customzone
  state: present
- set_default_zone: customzone
   state: enabled
 - service: [tftp, ftp]
   port: ['443/tcp', '443/udp']
   forward_port: ['447/tcp;;1.2.3.4',
                   '448/tcp;;1.2.3.5']
   state: enabled
- zone: internal
   service: [tftp, ftp]
   port: ['443/tcp', '443/udp']
   forward_port: ['447/tcp;;1.2.3.4',
                  '448/tcp;;1.2.3.5']
   state: enabled
- masquerade: true
- service: http
  state: enabled
```

Could you provide any instructions on how to replicate this error?

Thank you,

Brennan Paciorek

juliaschindler:
Hello!
I am experiencing this error for example using the following playbook against a freshly installed RHEL 9.2 host:

```
---
- hosts: all
  become: true
  gather_facts: false

  tasks:
    - name: Test firewall configuration
      vars:
        firewall:
          - previous: replaced
          - service: cockpit
            zone: public
            state: disabled
            permanent: true
      ansible.builtin.include_role:
        name: rhel-system-roles.firewall
```

Running the playbook without `--check` it completes without error.

PR #163, opened 2023-07-26
fix: error when running with check mode and previous: replaced

richm:
Cause: The role was not checking to see if any files would be
changed when using `previous: replaced` in check mode.  The variables
that should have been set were not set in check mode.

Consequence: The role would give an error about undefined variables.

Fix: A new check has been added for check mode, to see if any files
would be changed by `previous: replaced`.  The check for
`firewalld.conf` looks at the rpm database to determine if the file
has been changed from the version shipped in the package.  If files
would not be changed by using `previous: replaced`, then the role
will also now correctly report if any settings would be changed.

Result: The role works in check mode when using `previous: replaced`

### DIFF
--- a/files/get_files_checksums.sh
+++ b/files/get_files_checksums.sh
@@ -5,6 +5,8 @@ set -euo pipefail
 python_cmd="$1"
 firewall_conf_root="${2:-/etc/firewalld}"
 remove="${3:-false}"
+package="${4:-}"
+firewall_usr_lib="${5:-}"
 
 listfile=$(mktemp)
 firewallconf=$(mktemp)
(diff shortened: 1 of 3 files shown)


tasks/main.yml
```diff
@@ -49,6 +49,7 @@
   register: __firewall_config_files_before
   changed_when: false
   when: __firewall_previous_replaced | bool
+  check_mode: false
```
BrennanPaciorek, line 52: While this would prevent errors, I think disabling check mode for this task would be a bad idea if we want to support check mode with `previous: replaced`, or at least it is insufficient on its own.

In order to make this a full solution, I'd recommend modifying files/get_files_checksums.sh to support check mode when `remove=true`

richm: What should we report when the user runs this with check mode?

```yaml
- name: Run role
  include_role:
  vars:
    firewall:
      - previous: replaced
```

That is - how do we know that the firewall configuration is already in the "default" state?

richm:
[citest]

tasks/main.yml
```diff
@@ -40,16 +40,60 @@
...
+  block:
+    - name: Get config files, checksums before and remove
+      script:
+        cmd: >
+          files/get_files_checksums.sh {{ __firewall_python_cmd | quote }}
+          {{ __firewall_firewalld_dir | quote }} true
+          {{ __firewall_package_with_conf }}
+      register: __firewall_config_files_before
+      changed_when: false
+      when: not ansible_check_mode
```
BrennanPaciorek, line 56: I think the ansible.builtin.script will refuse to run in check mode anyways without `check_mode: false` or either `removes` or `creates` set with the error msg:
"Check mode is not supported for this task."

https://github.com/ansible/ansible/blob/010e542a13516eed9bb6b5e997ea255104257a80/lib/ansible/plugins/action/script.py#L100

Keeping `not ansible_check_mode` Is  probably the best choice, since we cannot guarantee that ansible will keep this behavior in later releases.

tasks/main.yml
```diff
@@ -40,16 +40,60 @@
...
+
+    - name: Check config files to see if they would be changed
+      find:
+        paths:
+          - "{{ __firewall_firewalld_dir }}"
+        patterns:
+          - "*.xml"
+      register: __firewall_files_found
+      changed_when: __firewall_files_found.matched > 0
+      when: ansible_check_mode | bool
```
BrennanPaciorek, line 66: NetworkManager integration makes some changes, so we'll need to find a workaround for that if there is anything executing firewall-cmd after this task

files/get_files_checksums.sh
```diff
@@ -16,21 +17,29 @@ find "$firewall_conf_root" -name \*.xml | while read -r file; do
...
+    else
+        remove_firewall_conf=false
+    fi
 fi
 
 if [ "${remove:-false}" = true ]; then
     find "$firewall_conf_root" -name \*.xml -exec rm -f {} \;
-    rm -f "$firewall_conf_root/firewalld.conf"
+    if [ "$remove_firewall_conf" = true ]; then
+        rm -f "$orig_conf"
```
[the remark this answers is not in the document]
richm: maybe - let's see what the latest commit does

tasks/main.yml
richm: I rewrote this somewhat

tasks/main.yml
richm: I now compare the checksum of the file under /etc/firewalld with the file under /usr/lib/firewalld

richm:
[citest]

richm:
ok - tests look good - I can verify locally that check mode appears to be working

for testing check mode:
* `tox -e qemu-ansible-core-2.15 -- --image-name centos-9 --log-level debug --debug -- tests/tests_default.yml`

this will ensure packages and services are present and leave the vm running

* `ansible-playbook -vv -i /tmp/inventory-cloudxxxxx/inventory --check tests/tests_check_mode.yml`

find the tmp inventory, then use that to run ansible-playbook against the running vm using check mode

BrennanPaciorek:
Works well for me but a test like this fails with the same error as in #151:
```yaml
---
- hosts: all
  become: true
  gather_facts: false

  tasks:
    - name: Previous replaced in check mode
      vars:
        firewall:
      ansible.builtin.include_role:

    - name: Enter check mode
      check_mode: true
      block:
        - name: Previous replaced
          ansible.builtin.include_role:
          vars:
            firewall:
              - previous: replaced
```

This test seems to fail, making a suggestion now to fix that

tasks/main.yml
```diff
@@ -141,13 +153,17 @@
...
     - name: Get config files, checksums after
       script:
         cmd: >
           files/get_files_checksums.sh {{ __firewall_python_cmd | quote }}
-          {{ __firewall_firewalld_dir | quote }}
+          {{ __firewall_firewalld_dir | quote }} false
+          {{ __firewall_package_with_conf | quote }}
+          {{ __firewall_usr_lib_dir | quote }}
       register: __firewall_config_files_after
       changed_when: false
```
BrennanPaciorek, line 169:
```suggestion
          {{ __firewall_firewalld_dir | quote }} false
          {{ __firewall_package_with_conf | quote }}
          {{ __firewall_usr_lib_dir | quote }}
      register: __firewall_config_files_after
      changed_when: false
      check_mode: false
```
Without removing check mode, the script will not run which will result in failure on task `Calculate what has changed`

richm: I don't think using `check_mode: true` on the block is the same as running the playbook with `--check` - the problem is that I do not want the block starting at line 162 to be executed *at all* in check mode - this works correctly when running the playbook with `--check` - but when using `check_mode: true`, somehow `ansible_check_mode` is both true and false - it is False because the block is executed, but it is True because the script task complains . . .

BrennanPaciorek (approved):

Issue #106, opened 2022-11-08
creating ipset


sugitk:
**What would you like to be added**:

I would like to run the command like "firewall-cmd --new-ipset=foobar --permanent --type=hash:ip" in a playbook rather than using a command module.

**Why is this needed**:

The customer requested in the technical support channel.

richm:
Can any of you try out the proposed PR?

PR #166, opened 2023-07-28
feat: define, modify, and remove ipsets

BrennanPaciorek:
User can specify `state: present` or `state: absent` and `permanent: true` with new ipset arguments to configure ipsets for use in zones using the `source` argument
- firewall_lib.py
  - new argument: ipset - name of ipset
  - new argument: ipset_type - type of ipset
  - new argument: ipset_entry - contents of ipset
  - protections against failure in check mode when enabling and disabling ipsets for zones
- new file: tests/tests_ipsets.yml
  - tests user defined ipsets (create, modify, delete, use)
- tests: unit: new test cases for triggering ipset warnings and errors
- docs: README, firewall_lib DOCUMENTATION for ipset feature

Enhancement:
Users can define, modify, and delete ipsets using the system role, which can be added to and removed from zones
or be used when defining rich rules.

Reason:
IPSets make firewalld configuration much easier to maintain:
- Rich rules defining rules for many IP addresses can be made much smaller
- Allows for semantic grouping of IP addresses

Also, brings the srole closer to being a full solution for managing firewalld configuration.

Result:
Users should be able to manage ipsets using the firewall system role using the following arguments:
- `ipset`
- `ipset_type`
- `ipset_entries`
- `short`
- `description`
- `state: present` or `state: absent`
- `permanent: true`

GitHub Issue #106 
BZ 2140880 - https://bugzilla.redhat.com/show_bug.cgi?id=2140880


library/firewall_lib.py
```diff
@@ -1138,6 +1216,83 @@
...
+                            % (ipset_source_name, bound_zones)
+                        )
+                    else:
+                        module.fail_json(
+                            msg="Remove %s from all permanent and runtime zones before attempting to remove it"
+                            % ipset_source_name
+                        )
+                else:
+                    if not module.check_mode:
+                        fw_ipset.remove()
```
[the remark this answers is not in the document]
BrennanPaciorek: This code should not be reachable if when it is not initialized.

BrennanPaciorek:
TODO: increase unit test patch coverage.

README.md
```diff
@@ -364,6 +428,36 @@ port: '443/tcp'
 port: ['443/tcp','443/udp']
 ```
 
+### ipset_type
+
+Type of ipset being defined.
+Used with `ipset`.
+
+For a list of available ipset types, run `firewall-cmd --get-icmp-types`,
```
richm, line 436:
```suggestion
For a list of available ipset types, run `firewall-cmd --get-ipset-types`,
```

library/firewall_lib.py
```diff
@@ -317,6 +342,22 @@ def create_service(module, fw, service):
...
+    if not module.check_mode:
+        fw_ipset_settings = FirewallClientIPSetSettings()
+        fw_ipset_settings.setType(ipset_type)
+        fw.config().addIPSet(ipset, fw_ipset_settings)
+        fw_ipset = fw.config().getIPSetByName(ipset)
+        fw_ipset_settings = fw_ipset.getSettings()
+    else:
+        fw_ipset = None
+        fw_ipset_settings = FirewallClientIPSetSettings()
+        fw_ipset_settings.setType(ipset_type)
```
richm, line 357: Does this actually change anything in check mode?

library/firewall_lib.py
```diff
@@ -777,9 +827,14 @@ def main():
                 " port, source_port, protocol, destination, or helper_module)"
             )
 
-        # Zone and service are incompatible when state is set to present or absent
-        if zone and service:
-            module.fail_json(msg="both zone and service while state present/absent")
+        # Zone, service, and ipset are incompatible when state is set to present or absent
+        num_conflicting_args = len([x for x in [zone, ipset, service] if x])
```
richm, line 831: I think the code would be much simpler if you used `any` https://docs.python.org/2/library/functions.html#any - I don't think we need to know the number of conflicting args or which ones conflict - I think it is sufficient to say "more than one of zone, ipset, or service was specified with present/absent" or something like that

```python
if any((zone, ipset, service)): ... fail ...
```

library/firewall_lib.py
```diff
@@ -1140,6 +1213,85 @@ def exception_handler(exception_message):
...
+                and description != fw_ipset_settings.getDescription()
+            ):
+                if not module.check_mode:
+                    fw_ipset_settings.setDescription(description)
+                changed = True
+            if short is not None and short != fw_ipset_settings.getShort():
+                if not module.check_mode:
+                    fw_ipset_settings.setShort(short)
+                changed = True
+            for entry in ipset_entries:
```
richm, line 1246: what if the list of `ipset_entries` is smaller than the existing list of ipset entries for this ipset?  should the code remove entries that aren't in the given `ipset_entries`?

BrennanPaciorek: with `state: present`, `ipset_entries` will add the entry if it is not in the ipset, and do nothing otherwise. If 
`state: absent`, `ipset_entries` will remove the entries if they are present in the ipset, and do nothing if they are not present.

Would it be better if we simply had `ipset_entries` replace the entries in an ipset when specified with `state: present`, then not work with `state: absent`?

library/firewall_lib.py
BrennanPaciorek: This avoids calling fw.config().addIPSet(), which adds the ipset to permanent configuration.

This organization does create some redundant code though, so I'll refactor to remove that redundancy.

library/firewall_lib.py
BrennanPaciorek: > ```python
> if any((zone, ipset, service)): ... fail ...
> ```

A conditional like this wouldn't work, since it would cause a failure if any of these were set, when the only fail case should be if two or more of these arguments. This is here mostly because I thought it was cleaner than checking for the other two arguments if `zone` is specified, then `ipset` if `service` is set

BrennanPaciorek: Also, updated the comments to be more clear that those three options are not incompatible with `state: present` and `state: absent` but rather that they are incompatible with one another.

library/firewall_lib.py
richm: > with `state: present`, `ipset_entries` will add the entry if it is not in the ipset, and do nothing otherwise. If `state: absent`, `ipset_entries` will remove the entries if they are present in the ipset, and do nothing if they are not present.

It should be consistent with how other types work - so if services work like this, then ipset should also.

> 
> Would it be better if we simply had `ipset_entries` replace the entries in an ipset when specified with `state: present`, then not work with `state: absent`? This was one design that I had considered, but decided against due in favor of consistency with how services are defined, and how the firewall ansible_fact allows the user to see the full configuration of the ipset.

richm:
[citest]

richm:
[citest]

richm (approved):

PR #171, opened 2023-08-09
tests: test_ping: fix compatability issues

BrennanPaciorek:
Test script no longer fails if managed node does not support podman network create --interface name

Enhancement:
- Removed unnecessary imports from Containerfile generated by tests/files/test_ping.sh
- Removed usage of --interface-name from podman network create in test_ping.sh (and removed usage of podman network create in general, since that is also unnecessary for the purposes of the test)

Reason:
test fails on some platforms (RHEL 8.8, 9.2) due to podman not supporting `podman network create --interface-name` option

Result:
Tests pass on managed nodes with podman versions < v4.4

### DIFF
--- a/tests/files/test_ping.sh
+++ b/tests/files/test_ping.sh
@@ -11,25 +11,21 @@ cleanup() {
   rm -f /tmp/Containerfile
   podman stop --all
   podman rm --all
-  podman network rm podmanbr0 || :
 }
 trap "cleanup 1>/dev/null" EXIT
 
 cat > /tmp/Containerfile << EOF
 FROM quay.io/centos/centos:stream8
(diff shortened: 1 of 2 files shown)


BrennanPaciorek:
@richm These changes seem to fix the issue on the affected versions when tested on my machine, can you test out the PR as well?

richm:
> @richm These changes seem to fix the issue on the affected versions when tested on my machine, can you test out the PR as well?

lgtm - I added some debugging to make it easier to diagnose issues with this script in the future - https://github.com/BrennanPaciorek/firewall/pull/1

richm:
[citest]

richm (approved):

Issue #164, opened 2023-07-27
Overwite /etc/firewalld.conf on previous replaced instead of deleting it


BrennanPaciorek:
**What would you like to be added**:

Overwriting of changed directives in /etc/firewalld.conf, instead of deleting firewalld.conf if it is different from the default configuration file.

**Why is this needed**:

When /etc/firewalld.conf is deleted, comments describing its directives are also removed (see issue #127).

richm:
@BrennanPaciorek is this done?  If so, please add a link to the PR and close it

PR #176, opened 2023-08-15
fix: files: overwrite firewalld.conf on previous replaced

BrennanPaciorek:
on previous replaced, revert firewalld.conf to fallback configuration instead of deletion

Enhancement:
Replaces firewalld.conf on previous replaced instead of deleting when changed

Reason:
EL7+ firewalld releases appeared to support a fallback configuration in firewalld, which could be effectively used to override the previous configuration.

Result:
On reset, firewalld.conf is not deleted, and comments in the file are retained.

Addresses GitHub issue #164

### DIFF
--- a/files/get_files_checksums.sh
+++ b/files/get_files_checksums.sh
@@ -30,10 +30,12 @@ find "$firewall_conf_root" -name \*.xml | while read -r file; do
     fi
 done > "$listfile"
 
+set +o pipefail
+
 orig_conf="$firewall_conf_root/firewalld.conf"
 remove_firewall_conf=true
 if [ -f "$orig_conf" ]; then
-    if [ -z "$package" ] || (rpm -V "$package" || true) | grep -q "c ${orig_conf}$"; then
+    if [ -z "$package" ] || rpm -V "$package" | grep -q "c ${orig_conf}$"; then
         cp "$orig_conf" "$firewallconf"
         "$python_cmd" -c 'import os, sys
 from firewall.core.io.firewalld_conf import firewalld_conf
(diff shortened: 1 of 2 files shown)


files/get_files_checksums.sh
```diff
@@ -33,7 +33,7 @@ done > "$listfile"
 orig_conf="$firewall_conf_root/firewalld.conf"
 remove_firewall_conf=true
 if [ -f "$orig_conf" ]; then
-    if [ -z "$package" ] || rpm -V "$package" | grep -q "c ${orig_conf}$"; then
+    if [ -z "$package" ] || (rpm -V "$package" || true) | grep -q "c ${orig_conf}$"; then
```
richm, line 36: Is the problem here the `set -o pipefail` behavior?  You may have to turn it off e.g. on the line before this one do `set +o pipefail`, then after set it back to `set -o pipefail`

tests/tests_purge_config.yml
```diff
@@ -143,6 +143,27 @@
...
+
+        - name: Purge config
+          include_role:
+            name: linux-system-roles.firewall
+          vars:
+            firewall:
+              - previous: replaced
+
+        - name: Fail if /etc/firewalld/firewalld.conf no longer exists
+          command: test -f /etc/firewalld/firewalld.conf
```
richm, line 164: better to use `stat`, `register`, and `failed_when` here

```yaml
stat:
  path: /etc/firewalld/firewalld.conf
register: __stat
failed_when: not __stat.stat.exists
```

BrennanPaciorek:
Added some extra tests to verify that firewalld.conf won't change after running `previous: replaced` with no changes.

richm:
[citest]

richm (approved):
lgtm - @BrennanPaciorek is this PR ready, or do you have more changes planned?

BrennanPaciorek:
No additional changes planned. This is good to merge

PR #184, opened 2023-10-03
ci: add dump_packages to weekly_ci

richm:
Add dump_packages callback to weekly_ci so we can get
lists of packages for ostree builds.


richm:
[citest]

PR #191, opened 2023-10-10
feat: support for ostree systems

richm:
Feature: Allow running and testing the role with ostree managed nodes.

Reason: We have users who want to use the role to manage ostree
systems.

Result: Users can use the role to manage ostree managed nodes.


richm:
[citest]

richm:
@spetrosi any objection to merging?  I need to publish the roles used by other roles first

PR #226, opened 2024-07-03
feat: Handle reboot for transactional update systems

HVSharma12:
Enhancement: - Added a block to handle reboots for transactional update systems on firewall package changes. Update README.

Reason: Ensure necessary reboots are managed, as changes on transactional update systems are applied in a separate snapshot and require a reboot to take effect.

Result: Manages system reboots for transactional update systems when package is installed.

Issue Tracker Tickets (Jira or BZ if any):na

### DIFF
--- a/README.md
+++ b/README.md
@@ -771,12 +771,12 @@ being reset and all new connections to the system are rejected.  Existing
 connections will be unaffected. Applying changes with this option in production
 might cause temporary service failures with new connections during the operation.
 
-### transactional_update_reboot_ok
+### firewall_transactional_update_reboot_ok
 
-This variable is used to handle reboots required by transactional updates. If a transactional update requires a reboot, the role will proceed with the reboot if transactional_update_reboot_ok is set to true. If set to false, the role will notify the user that a reboot is required, allowing for custom handling of the reboot requirement. If this variable is not set, the role will fail to ensure the reboot requirement is not overlooked.
+This variable is used to handle reboots required by transactional updates. If a transactional update requires a reboot, the role will proceed with the reboot if firewall_transactional_update_reboot_ok is set to true. If set to false, the role will notify the user that a reboot is required, allowing for custom handling of the reboot requirement. If this variable is not set, the role will fail to ensure the reboot requirement is not overlooked.
 
 ```yaml
-transactional_update_reboot_ok: true
+firewall_transactional_update_reboot_ok: true
 ```
 
 ## Examples of Options
--- a/README.md
+++ b/README.md
@@ -771,6 +771,19 @@ being reset and all new connections to the system are rejected.  Existing
 connections will be unaffected. Applying changes with this option in production
 might cause temporary service failures with new connections during the operation.
 
+### system_is_transactional
+
+This variable indicates whether the system uses transactional updates. It should be set to true if the system supports and is configured for transactional updates.
+
+### transactional_update_reboot_ok
+
+This variable is used to handle reboots required by transactional updates. If a transactional update requires a reboot, the role will proceed with the reboot if transactional_update_reboot_ok is set to true. If set to false, the role will notify the user that a reboot is required, allowing for custom handling of the reboot requirement. If this variable is not set, the role will fail to ensure the reboot requirement is not overlooked.
+
+```yaml
+system_is_transactional: true
+transactional_update_reboot_ok: true
+```
+
 ## Examples of Options
 
 By default, any changes will be applied immediately, and to the permanent settings. If you want the changes to apply immediately but not permanently, use `permanent: false`. Conversely, use `runtime: false`.
--- a/README.md
+++ b/README.md
@@ -771,16 +771,11 @@ being reset and all new connections to the system are rejected.  Existing
 connections will be unaffected. Applying changes with this option in production
 might cause temporary service failures with new connections during the operation.
 
-### system_is_transactional
-
-This variable indicates whether the system uses transactional updates. It should be set to true if the system supports and is configured for transactional updates.
-
 ### transactional_update_reboot_ok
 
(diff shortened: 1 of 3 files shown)


tasks/firewalld.yml
```diff
@@ -23,3 +23,27 @@
...
+  when:
+    - system_is_transactional | d(false)
+    - firewall_package_result is changed
+  block:
+    - name: Notify user that reboot is needed to apply changes
+      debug:
+        msg: >
+          Reboot required to apply changes due to transactional updates.
+
+    - name: Fail if reboot is needed and not set
```
spetrosi, line 38: In other system roles, the role does not fail, but print a message that reboot is required and sets`storage_reboot_required: true` fact. See https://github.com/linux-system-roles/bootloader/blob/main/tasks/reboot.yml
However, here the role won't be able to proceed without the reboot. So I guess this is fine.

README.md
```diff
@@ -771,6 +771,19 @@ being reset and all new connections to the system are rejected.  Existing
 connections will be unaffected. Applying changes with this option in production
 might cause temporary service failures with new connections during the operation.
 
+### system_is_transactional
```
spetrosi, line 774: Variables must be prefixed with `storage_`

tasks/firewalld.yml
```diff
@@ -23,3 +23,27 @@
     state: present
     use: "{{ (__firewall_is_ostree | d(false)) |
       ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
+  register: firewall_package_result
+
+- name: Handle reboot for transactional update systems
+  when:
+    - system_is_transactional | d(false)
```
spetrosi, line 30: Define those variables in `defaults/main.yml`, set both to `false there. Then, the `| d` filter is not needed.

README.md
spetrosi: Is there some way to determine if a system is transactional?

HVSharma12: Would it be better to use transactional_update_ or __firewall_is_transactional as a prefix instead? This way, it will be clear that the variables and tasks are related to transactional update systems rather than storage systems. 

One way to determine would be to check for /sbin/transactional-update stat similar to the ostree tasks, or would it be better to keep it as user-defined? -

~~~
 - name: Determine if system is transactional update and set flag
      when: not __firewall_is_transactional is defined
      block:
        - name: Check if transactional-update exists in /sbin
          stat:
~~~
~~~
    - name: Handle reboot for transactional update systems
      when:
        - __firewall_is_transactional | d(false)
        - firewall_package_result is changed
~~~

tasks/firewalld.yml
HVSharma12: When setting the reboot variable in the defaults, the fail task is bypassed as it looks for 'is undefined'.
__
This variable is used to handle reboots required by transactional updates. If a transactional update requires a reboot, the role will proceed with the reboot if transactional_update_reboot_ok is set to true. If set to false, the role will notify the user that a reboot is required, allowing for custom handling of the reboot requirement. If this variable is not set, the role will fail to ensure the reboot requirement is not overlooked.

richm: We have a pretty strong preference for requiring all variables that are part of the public API of the role to be specified in defaults/main.yml with the default value.  In this case, the default value can be `null`. Then the test will be

```yaml
  when:
    - system_is_transactional is not none
    - system_is_transactional | d(false)
```

or if you prefer a one-liner:

```yaml
when: (system_is_transactional is none) | ternary(false, system_is_transactional)
```

We use this convention in many, many places in system roles.

README.md
richm: For example, the role determines if the system is an rpm-ostree based system by looking for the file /run/ostree-booted and setting some variables - https://github.com/linux-system-roles/firewall/blob/main/tasks/firewalld.yml#L13

But perhaps this case is different?  I know that on rpm-ostree systems, you can use the `rpm-ostree` command to stage packages for installation, then reboot the system in order to apply the changes.  Perhaps your transactional system is similar? and there is no way for the role to determine if the system is transactional?

tasks/firewalld.yml
```diff
@@ -17,9 +17,45 @@
...
+    - firewall_package_result is changed
+  block:
+    - name: Notify user that reboot is needed to apply changes
+      debug:
+        msg: >
+          Reboot required to apply changes due to transactional updates.
+
+    - name: Reboot transactional update systems
+      reboot:
+        msg: "Rebooting the system to apply transactional update changes."
```
spetrosi, line 52:
```suggestion
        msg: Rebooting the system to apply transactional update changes.
```

defaults/main.yml
```diff
@@ -1,3 +1,4 @@
 ---
 firewall: []
 firewall_disable_conflicting_services: false
+transactional_update_reboot_ok: null
```
spetrosi, line 4: > Would it be better to use transactional_update_ or __firewall_is_transactional as a prefix instead? This way, it will be clear that the variables and tasks are related to transactional update systems rather than storage systems.

ansible-lint requires role variables to be prefixed with role name. It's a good practice to group all role's variables for the sake of clarity. So this should be `firewall_transactional_update_reboot_ok` .

tasks/firewalld.yml
```diff
@@ -17,9 +17,45 @@
...
+  block:
+    - name: Notify user that reboot is needed to apply changes
+      debug:
+        msg: >
+          Reboot required to apply changes due to transactional updates.
+
+    - name: Reboot transactional update systems
+      reboot:
+        msg: Rebooting the system to apply transactional update changes.
+      when:
```
richm, line 53:
```suggestion
      when: firewall_transactional_update_reboot_ok | bool
```

tasks/firewalld.yml
```diff
@@ -17,9 +17,45 @@
...
+        - firewall_transactional_update_reboot_ok
```
richm, line 54:
```suggestion
```

richm (approved):

richm:
[citest]

Issue #234, opened 2024-08-19
forward_ports not persistent


rlenferink:
**What happened**:

I am using the firewall role with the following definition:

```yml
firewall:
  # Wipe existing config and replace with the config defined below.
  - previous: replaced

  # For some reason the cockpit and dhcpv6-client are enabled by default.
  # Explicitly disable them.
  - service:
      - cockpit
      - dhcpv6-client
    state: disabled

  # Only allow the ports below.
  - zone: public
    state: enabled
    permanent: true
    interface: ens18
    services:
      - ssh
    port:
        # Web ports
      - '80/tcp'    # HTTP
      - '443/tcp'   # HTTPS

    # NOTE: For some reason the forward_ports are not persistent. When making changes always make them persistent manually using:
    #   firewall-cmd --runtime-to-permanent
    masquerade: true
    forward_port:
      - port: 1194
        proto: udp
        toaddr: 192.168.1.1
        toport: 1195

      - port: 16001
        proto: tcp
        toaddr: 192.168.1.21  # nas.loc (Video Station HTTPS)
        toport: 16001
```

**What you expected to happen**:

I expected because of the `permanent` flag that the `forward_port`s would also be persistent. The rest of the firewall configuration is persistent (if no `forward_port`s are specified), however when forward_ports are specified, the configuration is lost after a `reboot` of the target node.

I verified this with `firewall-cmd --list-all-zones`. It seems that the forward_ports are gone, and if there is a forward_port specified that the 'normal' ports are also gone.

**How to reproduce it (as minimally and precisely as possible)**:

Use e.g. the snipped above.

**Anything else we need to know?**:

**Environment**:
- Firewalld Version (if Fedora based `dnf info firewalld` or commit hash if developing from git `git log -n1 --format=format:"%H"`):
- Firewalld Backend (`cat /etc/firewalld/firewalld.conf | grep FirewallBackend`): 0.9.11
- Ansible Version (`ansible --version`): 2.16.5
- Firewall System Role Version (if installing from ansible galaxy `ansible-galaxy list | grep linux-system-roles.firewall` or commit hash if developing from git
`git log -n1 --format=format:"%H"`): 4c4a1343ac0afc74e98d7c583e9258b29131fe55
- OS (e.g: `cat /etc/os-release`): Rocky Linux 8.10
- Others:
  - Ansible controller node is Fedora 40
  - This is my home environment

richm:
[...]

PR #237, opened 2024-10-15
fix: try to debug issue 234

richm:
Try to debug and reproduce https://github.com/linux-system-roles/firewall/issues/234


richm:
[citest]

PR #241, opened 2024-10-20
fix: Prevent interface definitions overriding 'changed' value when other elements are changed

### LINKED ISSUE #234: forward_ports not persistent

rlenferink:
Cause: The module was replacing the `changed` value with `False` when using `interface` and the interface exists.

Consequence: The role reported `changed: False` even when there were changes, and `forward_ports` were not being saved as permanent.

Fix: Ensure the `changed` value is not reset to `False` in this case.

Result: The role reports `changed: True` when there are changes, and `forward_ports` settings are persistent.

This fixes #234

### DIFF
--- a/library/firewall_lib.py
+++ b/library/firewall_lib.py
@@ -1469,7 +1469,8 @@ def main():
             if permanent:
                 nm_used, if_changed = try_set_zone_of_interface(module, zone, item)
                 if nm_used:
-                    changed = if_changed
+                    if if_changed:
+                        changed = True
                 elif not fw_settings.queryInterface(item):
                     if not module.check_mode:
                         handle_interface_permanent(
@@ -1484,7 +1485,8 @@ def main():
             if permanent:
                 nm_used, if_changed = try_set_zone_of_interface(module, "", item)
                 if nm_used:
-                    changed = if_changed
+                    if if_changed:
+                        changed = True
                 elif fw_settings.queryInterface(item):
                     if not module.check_mode:
                         fw_settings.removeInterface(item)
--- a/tests/unit/test_firewall_lib.py
+++ b/tests/unit/test_firewall_lib.py
@@ -785,7 +785,7 @@ class FirewallLibMain(unittest.TestCase):
     ):
         am = am_class.return_value
         am.params = {
-            "interface": "eth0",
+            "interface": ["eth0"],
             "zone": "public",
             "permanent": True,
         }
@@ -799,6 +799,13 @@ class FirewallLibMain(unittest.TestCase):
 
         fw_config = Mock()
         fw_config.getZoneNames.return_value = available_zones
+        fw_config.getServiceNames.return_value = ["ssh"]
+        fw_zone = Mock()
+        fw_config.getZoneByName.return_value = fw_zone
+        fw_settings = Mock()
+        fw_settings.queryService.return_value = False
+        fw_settings.addService = Mock()
+        fw_zone.getSettings.return_value = fw_settings
         fw.config.return_value = fw_config
 
         return_values = [(True, True), (True, False)]
@@ -806,8 +813,18 @@ class FirewallLibMain(unittest.TestCase):
             am.params["state"] = state
             for rv in return_values:
                 try_set_zone_of_interface.return_value = rv
+                fw_settings.queryService.return_value = state == "disabled"
                 firewall_lib.main()
                 am.exit_json.assert_called_with(changed=rv[1], __firewall_changed=rv[1])
+
+        for state in ["enabled", "disabled"]:
+            am.params["state"] = state
+            am.params["service"] = ["ssh"]
+            for rv in return_values:
+                try_set_zone_of_interface.return_value = rv
+                fw_settings.queryService.return_value = state == "disabled"
+                firewall_lib.main()
+                am.exit_json.assert_called_with(changed=True, __firewall_changed=True)
 
     @patch("firewall_lib.HAS_FIREWALLD", True)
     @patch("firewall_lib.FW_VERSION", "0.9.0", create=True)


library/firewall_lib.py
```diff
@@ -1468,8 +1468,8 @@ def exception_handler(exception_message):
                 changed = True
             if permanent:
                 nm_used, if_changed = try_set_zone_of_interface(module, zone, item)
-                if nm_used:
-                    changed = if_changed
+                if nm_used and if_changed:
+                    changed = True
```
richm, line 1472: We only want the `elif` clause executed if `nm_used` is `False`.  So try this instead:

```python
                if nm_used:
                    if if_changed:
                        changed = True
```
and same below

richm:
This will also require a fix in the unit tests.  This works for me:

```
diff --git a/tests/unit/test_firewall_lib.py b/tests/unit/test_firewall_lib.py
index 0497ad9..dea560c 100644
--- a/tests/unit/test_firewall_lib.py
+++ b/tests/unit/test_firewall_lib.py
@@ -785,7 +785,7 @@ class FirewallLibMain(unittest.TestCase):
     ):
         am = am_class.return_value
         am.params = {
-            "interface": "eth0",
+            "interface": ["eth0"],
             "zone": "public",
             "permanent": True,
         }
@@ -799,6 +799,13 @@ class FirewallLibMain(unittest.TestCase):
 
         fw_config = Mock()
         fw_config.getZoneNames.return_value = available_zones
+        fw_config.getServiceNames.return_value = ["ssh"]
+        fw_zone = Mock()
+        fw_config.getZoneByName.return_value = fw_zone
+        fw_settings = Mock()
+        fw_settings.queryService.return_value = False
+        fw_settings.addService = Mock()
+        fw_zone.getSettings.return_value = fw_settings
         fw.config.return_value = fw_config
 
         return_values = [(True, True), (True, False)]
@@ -809,6 +816,18 @@ class FirewallLibMain(unittest.TestCase):
                 firewall_lib.main()
                 am.exit_json.assert_called_with(changed=rv[1], __firewall_changed=rv[1])
 
+        # now try with some other change in the mix
+        # main should always return True due to the service change
+        # regardless of the zone interface change
+        return_values = [(True, True), (True, False)]
+        for state in ["enabled", "disabled"]:
+            am.params["state"] = state
+            am.params["service"] = ["ssh"]
+            for rv in return_values:
+                try_set_zone_of_interface.return_value = rv
+                firewall_lib.main()
+                am.exit_json.assert_called_with(changed=True, __firewall_changed=True)
+
     @patch("firewall_lib.HAS_FIREWALLD", True)
     @patch("firewall_lib.FW_VERSION", "0.9.0", create=True)
     def test_ipset_operation_with_services_set(self, am_class):
```

You helped me find a bug in the test - should be `"interface": ["eth0"]` instead of `"interface": "eth0"` - the only reason it worked before is because the string value is an iterable.

rlenferink:
@richm thanks for thinking along! I updated the pull request with your suggested change and the updated unit test.

tests/unit/test_firewall_lib.py
```diff
@@ -809,6 +816,18 @@ def test_nm_integration_interfaces(
                 firewall_lib.main()
```
richm, line 816:
```suggestion
                fw_settings.queryService.return_value = state == "disabled"
                firewall_lib.main()
```
sorry - looks like I previously posted the wrong diff :-(

tests/unit/test_firewall_lib.py
```diff
@@ -809,6 +816,18 @@ def test_nm_integration_interfaces(
...
 
+        # now try with some other change in the mix
+        # main should always return True due to the service change
+        # regardless of the zone interface change
+        return_values = [(True, True), (True, False)]
+        for state in ["enabled", "disabled"]:
+            am.params["state"] = state
+            am.params["service"] = ["ssh"]
+            for rv in return_values:
+                try_set_zone_of_interface.return_value = rv
```
richm, line 827:
```suggestion
                try_set_zone_of_interface.return_value = rv
                fw_settings.queryService.return_value = state == "disabled"
```

tests/unit/test_firewall_lib.py
```diff
@@ -809,6 +816,18 @@ def test_nm_integration_interfaces(
                 firewall_lib.main()
                 am.exit_json.assert_called_with(changed=rv[1], __firewall_changed=rv[1])
 
+        # now try with some other change in the mix
+        # main should always return True due to the service change
+        # regardless of the zone interface change
+        return_values = [(True, True), (True, False)]
```
richm, line 822:
```suggestion
```
This is already set above

richm:
@rlenferink sorry - I posted an incomplete diff earlier - I have made the suggested changes to the code - with these changes the python unit tests should pass

rlenferink:
@richm no problem! I processed your feedback.

richm:
[citest]

PR #258, opened 2025-04-08
ci: several changes related to new qemu test, ansible-lint, python versions, ubuntu versions

richm:
There is a new QEMU based test which uses the qemu/kvm capability of
github action runners.  This is the basis for new bootc/image mode tests
which we will be rolling out in the near future.

ansible-lint requires that the collection path is set so that the requirements
it installs are installed in the correct place.

There has been some general github action deprecation of python versions and
ubuntu versions that we have had to fix.

Remove `CONTRIBUTOR` from the list of users who can trigger citest.

For more information, see

* https://github.com/linux-system-roles/.github/pull/98
* https://github.com/linux-system-roles/.github/pull/94
* https://github.com/linux-system-roles/.github/pull/93
* https://github.com/linux-system-roles/.github/pull/92
* https://github.com/linux-system-roles/.github/pull/91


martinpitt:
I'm looking at the [centos-9 failure](https://github.com/linux-system-roles/firewall/actions/runs/14392660023/job/40362698527?pr=258) first. This is some shallow broken code, fixed in https://github.com/linux-system-roles/tox-lsr/pull/184 . But beyond cleaning this up, running this locally works fine. The actual HTTPError was most likely just a temporary network glitch on the CentOS compose mirror, and the download [worked in a retry](https://github.com/linux-system-roles/firewall/actions/runs/14392660023/job/40390706010?pr=258).

The `tests_zone` failure easily reproduces locally in tox, I'll investigate and create a reproducer for a bug report.

Issue #255, opened 2025-02-17
Add `includes` when defining a custom service


skwde:
**What would you like to be added**:

Allow for `includes` when defining a custom service (https://firewalld.org/documentation/man-pages/firewalld.service.html), e.g.

```yml
        - name: Run firewall role
          ansible.builtin.include_role:
            name: fedora.linux_system_roles.firewall
          vars:
            firewall_disable_conflicting_services: true
            firewall:
              - service: custom
                short: Custom
                description: Custom service
                port: 2222/tcp
                includes: [http, https]
                state: present
                permanent: true
```

Currently `includes` are silently ignored.

An example is e.g. the predefined firewalld service `freeipa-4` (https://github.com/firewalld/firewalld/blob/main/config/services/freeipa-4.xml)

**Why is this needed**:

This makes firewalld services more explicit and easier / quicker to read when there are many non-standard ports.

richm:
@skwde can you try https://github.com/linux-system-roles/firewall/pull/259 ?

PR #259, opened 2025-04-08
feat: support includes for services

richm:
Feature: Allow for includes to be specified for services.
Includes are described at https://firewalld.org/documentation/man-pages/firewalld.service.html

Reason: This makes firewalld services more explicit and easier / quicker
to read when there are many non-standard ports.

Result: Users can specify other services to include when creating
and setting services.

Fixes https://github.com/linux-system-roles/firewall/issues/255

### DIFF
--- a/README.md
+++ b/README.md
@@ -698,6 +698,19 @@ helper_module: ftp
 helper_module: nf_conntrack_ftp
 ```
 
+### includes
+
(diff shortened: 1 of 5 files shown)


richm:
[citest]

richm:
[citest]

PR #264, opened 2025-04-25
ci: Add container integration test for rpm and bootc

martinpitt:
Same as in https://github.com/linux-system-roles/sudo/pull/52 to check the new mechanics for a more complex role. This is a case where the role works fine in a system container, but *not* yet during a bootc container build. I'm working on that in my https://github.com/martinpitt/lsr-firewall/tree/bootc-container-test branch, see https://issues.redhat.com/browse/RHEL-88425 . But this is a nice stepping stone.

 - [x] https://github.com/linux-system-roles/tox-lsr/pull/191
 - [x] Release tox-lsr and bump to new release, then remove the extra XXX commit
 - [x] Review/land https://github.com/linux-system-roles/sudo/pull/52 as the initial PR for this machinery

### DIFF
--- a/tests/tests_reload_on_reset.yml
+++ b/tests/tests_reload_on_reset.yml
@@ -21,9 +21,12 @@
           set_fact:
             __firewall_is_ostree: "{{ __ostree_booted_stat.stat.exists }}"
 
-    - name: Install podman
+    - name: Install test dependencies
       package:
-        name: podman
+        name:
+          - podman
+          # for ping
+          - iputils
         state: present
         use: "{{ (__firewall_is_ostree | d(false)) |
                 ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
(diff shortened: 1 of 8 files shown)


martinpitt:
I.e. the [bootc scenarios](https://github.com/linux-system-roles/firewall/actions/runs/14665591594/job/41159486391?pr=264) are quick no-ops due to not being supported here.

tests/tests_reload_on_reset.yml
```diff
@@ -21,9 +21,11 @@
           set_fact:
             __firewall_is_ostree: "{{ __ostree_booted_stat.stat.exists }}"
 
-    - name: Install podman
+    - name: Install test dependencies
       package:
-        name: podman
+        name:
+          - podman
+          - /usr/bin/ping
```
richm, line 28: This introduces a dependency at ostree build time - either some package needs to be added to https://github.com/linux-system-roles/firewall/blob/main/.ostree/packages-testing.txt or possibly one of the platform specific files.  I don't know if image builder can use a file, or if it requires the name of a package.

martinpitt: Ack! Done.

richm (approved):

PR #266, opened 2025-04-29
test: skip include tests on el7, document el7 support

richm:
The new `includes` feature is only supported on EL8 and above, not
on EL7.  Document this.  Skip testing includes on EL7.


richm:
[citest]

PR #270, opened 2025-05-13
ci: Two prerequisites for bootc support

martinpitt:
These are two prerequisites for eventual bootc build support. They are [tested](https://github.com/martinpitt/lsr-firewall/actions/workflows/qemu-kvm-integration-tests.yml) in my [bootc-container-test branch](https://github.com/martinpitt/lsr-firewall/tree/refs/heads/bootc-container-test), and so far these parts work fine. These two commits are fairly intrusive but conceptually simple, so it may make sense to already land them. I also want to make sure that it doesn't break any of the qemu/TF tests.

## Summary by Sourcery

Prepare for bootc build support by migrating tests to offline mode and enhancing service handling.

Enhancements:
- Switch all ansible test scripts to use firewall-offline-cmd instead of firewall-cmd for offline configuration
- Replace service_facts with systemctl is-enabled checks and split disable/stop of conflicting services into separate tasks with conditional execution
- Wrap non-idempotent --set-default-zone calls in idempotent shell logic within tests

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -2,23 +2,24 @@
 - name: Setup firewalld
   include_tasks: firewalld.yml
 
-- name: Collect service facts
-  service_facts:
+- name: Check which conflicting services are enabled
+  # noqa command-instead-of-module
+  command: systemctl is-enabled "{{ item }}"
+  register: __firewall_conflicting_services_status
+  changed_when: false
+  failed_when: false
+  loop: "{{ __firewall_conflicting_services }}"
   when: firewall_disable_conflicting_services | bool
 
 - name: Attempt to stop and disable conflicting services
   service:
-    name: "{{ item }}"
+    name: "{{ item.item }}"
     state: stopped
     enabled: false
-  loop: "{{ __firewall_conflicting_services }}"
-  vars:
-    __service_name: "{{ (ansible_facts.service_mgr == 'systemd') |
-      ternary(item ~ '.service', item) }}"
+  loop: "{{ __firewall_conflicting_services_status.results }}"
   when:
     - firewall_disable_conflicting_services | bool
-    - __service_name | string in ansible_facts.services
-    - ansible_facts.services[__service_name]["status"] == "enabled"
+    - item.rc == 0
 
 - name: Unmask firewalld service
   systemd:
--- a/tests/tests_ansible.yml
+++ b/tests/tests_ansible.yml
@@ -9,7 +9,7 @@
         name: linux-system-roles.firewall
 
     - name: Get default zone
-      command: firewall-cmd --get-default-zone
+      command: firewall-offline-cmd --get-default-zone
       register: __default_zone
       changed_when: false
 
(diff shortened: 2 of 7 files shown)


martinpitt:
[citest]

martinpitt:
[citest]

martinpitt:
[citest]

tasks/main.yml
```diff
@@ -2,23 +2,32 @@
 - name: Setup firewalld
   include_tasks: firewalld.yml
 
-- name: Collect service facts
-  service_facts:
```
richm, line 6: Looks like we cannot use service_facts - a quick grep shows that we use it in 25 different places in our roles.

Could we use `systemctl list-unit-files --no-pager --type service --all` to somehow scan for unit files, then is-enabled on those?

We might need our own system roles `lsr_service_facts` module that we could use with all of our roles.

tasks/main.yml
```diff
@@ -2,23 +2,32 @@
...
+  changed_when: false
+  failed_when: false
+  loop: "{{ __firewall_conflicting_services }}"
   when: firewall_disable_conflicting_services | bool
 
-- name: Attempt to stop and disable conflicting services
+- name: Disable conflicting services
   service:
-    name: "{{ item }}"
-    state: stopped
```
richm, line 12: You could use a single task for this - and do something like this:

```yaml
  service:
    name: "{{ item }}"
    state: "{{ 'stopped' if __firewall_is_booted else omit }}"
...
```

richm:
I'm not sure how this works to simply swap `firewall-cmd` with `firewall-offline-cmd`, unless all of our tests always write the changes to the backing XML files.

martinpitt:
> I'm not sure how this works to simply swap `firewall-cmd` with `firewall-offline-cmd`, unless all of our tests always write the changes to the backing XML files.

The only places that use `runtime: true` are tests_ansible.yml and tests_ipsets.yml. But at these places they don't actually check the results. If they would, or we ever get a test that only covers `runtime: true`, we'd of course keep `firewall-cmd` and guard it with `when: __firewall_is_booted`.

All other places have `permanent: true` and called `firewall-cmd` with `--permanent` anyway, i.e. it's functionally equivalent.

tasks/main.yml
martinpitt: Ah nice! I wasn't aware of `omit` (I did ask ChatGPT, but that only suggested using `undefined()`, which doesn't work). I did that same role split in https://github.com/linux-system-roles/logging/commit/8b6916a7 , I sent https://github.com/linux-system-roles/logging/pull/448 to clean this up.

I will just drop that part here, as that condition will be part of the future PR.

tasks/main.yml
martinpitt: I considered `systemctl list-unit-files --no-legend --type service`, initially, but I found the rather elaborate string parsing and making assumptions about the output format more brittle than using a direct `is-enabled`  test -- that will only happen at most three times, so it's not very expensive. I can rewrite it if you feel strongly about it, of course.

tasks/main.yml
martinpitt: Done.

richm:
[citest]

tasks/main.yml
martinpitt: Yes, agreed -- if it's more than just a handful, then a summary `list-unit-files` is better. I hope not many roles need that, though.

richm (approved):

PR #273, opened 2025-05-19
tests: Various fixes and additions

martinpitt:
See individual commits for details. No runtime impact. These are just some issues I noticed while creating an offline backend.


martinpitt:
[citest]

martinpitt:
Added one more test.

richm (approved):

PR #277, opened 2025-05-20
fix: Fix "helpers" service option

martinpitt:
Cause: The query check was inverted, and query/add used the wrong API (module, not helper).

Consequence: Trying to set the "helpers:" option of a created service had no effect.

Fix: Use the correct API and fix the check. Add an integration test which actually validates the created service in firewalld.

Fixes #276

## Summary by Sourcery

Fix the helpers service option handling in firewall_lib by using the correct query and API methods so that helper modules are effectively applied, and add integration tests to validate custom service settings in firewalld.

Bug Fixes:
- Correct inverted helper query check and switch to queryHelper/addHelper calls for the helpers service option.

Tests:
- Add integration tests using firewall-offline-cmd to verify custom service settings including ports, protocols, destinations, and helper modules.

PR #278, opened 2025-05-20
fix: Fix "interface_pci_id" role option

martinpitt:
Cause: The introduction of this feature in commit 2b0bb1b9fb was broken: The new `interface_pci_id` option was not passed to `firewall_lib`. The integration test case was very shallow and did not verify the success of the operation.

Consequence: Calling the role with `interface_pci_id` had no effect, that setting was ignored.

Fix: Pass the option to the module. Rework the test to add a positive assertion that the interface mapped to the given PCI ID actually appears in the nftables/iptables rules.

Also replace the hardcoded 1af4:0001 ID in the test with runtime detection. That *happens* to be the ID of QEMU's virtio card, but this is neither obvious nor reliable nor does it apply to other environments such as Testing Farm.

Fixes #277


martinpitt:
Bwah, the firewall backend detection is utterly broken: On newer releases /usr/bin/dbus-send is not available any more (dbus-broker, `busctl`), installing it is hard as it gets shipped in different package names. One can install with `package: name: /usr/bin/dbus-send`, but then it's *still* broken on RHEL 7:

```
fatal: [/var/home/martin/.cache/linux-system-roles/centos-7.qcow2c]: FAILED! => {"changed": false, "cmd": ["dbus-send", "--system", "--print-reply", "--type=method_call", "--dest=org.fedoraproject.FirewallD1", "/org/fedoraproject/FirewallD1/config", "org.freedesktop.DBus.Properties.Get", "string:org.fedoraproject.FirewallD1.config", "string:FirewallBackend"], "delta": "0:00:00.007885", "end": "2025-05-20 12:51:42.454596", "msg": "non-zero return code", "rc": 1, "start": "2025-05-20 12:51:42.446711", "stderr": "Error org.freedesktop.DBus.Python.dbus.exceptions.DBusException: org.freedesktop.DBus.Error.InvalidArgs: Property 'FirewallBackend' does not exist", "stderr_lines": ["Error org.freedesktop.DBus.Python.dbus.exceptions.DBusException: org.freedesktop.DBus.Error.InvalidArgs: Property 'FirewallBackend' does not exist"], "stdout": "", "stdout_lines": []}
```

It's probably better for the test to just hardcode which OSes still expect iptables.

martinpitt:
[citest]

tests/tests_interface_pci.yml
```diff
@@ -7,32 +7,46 @@
...
 
     - name: Test interfaces with PCI ids
+      # this can't be tested in containers or similar envs without any real
+      # ethernet devices
+      when: find_iface.stdout != ""
       block:
+        - name: Determine interface vendor/product ID
+          shell: |
+            set -euo pipefail
+            VID="$(sed 's/^0x//' < {{find_iface.stdout}}/device/vendor)"
```
richm, line 37:
```suggestion
            VID="$(sed 's/^0x//' < {{ find_iface.stdout | quote }}/device/vendor)"
```
ansible wants a single space after `{{` and a single space before `}}`
quote ensures the value can be safely passed to shell

tests/tests_interface_pci.yml
```diff
@@ -7,32 +7,46 @@
...
+            PID="$(sed 's/^0x//' < {{find_iface.stdout}}/device/device)"
```
richm, line 38:
```suggestion
            PID="$(sed 's/^0x//' < {{ find_iface.stdout | quote }}/device/device)"
```

tests/tests_interface_pci.yml
```diff
@@ -7,32 +7,46 @@
...
+          changed_when: false
+
         - name: Add pci device ethernet controller
           include_role:
             name: linux-system-roles.firewall
           vars:
             firewall:
               zone: internal
-              interface_pci_id: 1af4:0001
+              interface_pci_id: "{{pci_id.stdout}}"
```
richm, line 49:
```suggestion
              interface_pci_id: "{{ pci_id.stdout }}"
```

tests/tests_interface_pci.yml
```diff
@@ -42,31 +56,43 @@
...
```
richm, line 59:
```suggestion
              interface_pci_id: "{{ pci_id.stdout }}"
```

tests/tests_interface_pci.yml
```diff
@@ -42,31 +56,43 @@
...
+          when: not nftables_backend | bool
 
         - name: Remove interface from internal
           include_role:
             name: linux-system-roles.firewall
           vars:
             firewall:
               zone: internal
-              interface_pci_id: 1af4:0001
+              interface_pci_id: "{{pci_id.stdout}}"
```
richm, line 95:
```suggestion
              interface_pci_id: "{{ pci_id.stdout }}"
```

richm (approved):

PR #310, opened 2025-12-04
refactor: ensure that using previous replaced does not remove files and reload unless necessary

richm:
The old implementation of `previous: replaced` would remove all files and reload the
firewall every time, even if not necessary - it wasn't technically idempotent, although
it would check the new files to see if they matched the old files, and report
`changed: false` if nothing actually changed state.

The new implementation uses an in-memory backend to apply the changes, then checks if
anything changed, and then removes the files and reloads firewall only if something
actually changed.


richm:
[citest]

PR #355, opened 2026-06-04
fix: ensure firewall_config is reset

richm:
Cause: The variable firewall_config is not being reset in the case where it was previously
requested, and the user runs the role again without requesting facts.

Consequence: firewall_config is set to the old value, so if the user uses it, they will get
the facts from the previous run.

Fix: Ensure that firewall_config is always set or reset by the role.

Result: The user will either get the current facts from firewall_config, or firewall_config
will be empty after a reset, and the user will not accidentally be using old facts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved firewall configuration initialization to ensure proper state reset before gathering configuration facts.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -79,6 +79,10 @@
         - firewall_lib_result.short_circuit is defined
         - not firewall_lib_result.short_circuit | bool
 
+    - name: Reset firewall_config
+      set_fact:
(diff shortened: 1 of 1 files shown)


richm:
@Timdawson264 fyi

spetrosi (approved):

Issue #177, opened 2023-08-16
Is "--add-protocol" for a zone not implemented?


Markus-:
**What would you like to be added**:
I want to manage des protocols of a zone by using this role.

`#> firewall-cmd --zone=block --add-protocol=ipv6-icmp`

**Why is this needed**:
For Ipv6 I want to add ipv6-icmp in an easy way.
https://firewalld.org/documentation/zone/options.html#protocol

Markus-:
I think it is undocumented, but possible. It is in the lib, but not in the readme.md.

I just tested it and it works.

Should we update the Documentation or did I miss the part about it?

Regards
Markus

Timdawson264:
There is no direct way to attach protocols to a zone with this role, but it can be done via a service.
Make a service that contans the protocol you wish to add, then attach the service to the zone. 
e.g:
```
- service: custom-ospfv3
  state: present
  short: OSPFv3 Service
  description: OSPFv3 Service 
  protocol: 89 #OSPFv3
```

Markus-:
Ok. Seems to be the workaround to use. Thanks.

PR #356, opened 2026-06-04
docs: explain how to add protocol to zone

richm:
Explain how to add a protocol to a zone by first defining a service with that protocol,
then adding the service to the zone.

## Summary by CodeRabbit

* **Documentation**
  * Enhanced zone configuration documentation with clarification that protocols must be attached via services rather than directly to zones.
  * Added example playbook snippet demonstrating the recommended protocol attachment workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,20 @@ variable with no other variables, and use `state: present` to add the zone, or
 zone: public
 ```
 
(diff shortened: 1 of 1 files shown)


richm:
addresses https://github.com/linux-system-roles/firewall/issues/177

PR #358, opened 2026-06-08
fix: ensure rich rule handling is idempotent

### LINKED ISSUE #357: Rich rules always report `changed` — `rules_str` vs `rich_rules` key mismatch
Hi — I think I've hit a false-positive `changed` for any zone that has rich
rules. It's cosmetic (the firewall state doesn't actually drift), but it breaks
idempotency reporting and trips CI / `--check` workflows that expect
`changed: false` on a reconverge.

## What happens

Re-running the role with an unchanged rich rule reports `changed: true` every
time. The permanent and runtime config are byte-identical across runs
(`firewall-cmd --list-rich-rules`, and `--check-config` stays clean), so nothing
is actually added or removed — it's purely the change accounting.

## Root cause

The in-memory backend reads existing rich rules under one dict key, but the
config reader writes them under another:

- `module_utils/firewall_lsr/get_config.py` stores a zone's rich rules as
  `element_settings["rules_str"] = element.getRichRules()`
- `library/firewall_lib.py` `set_rich_rule()` looks them up as
  `zone_config.get("rich_rules", [])`

Since `rich_rules` is never populated, that lookup is always `[]`, so every
requested rich rule looks new → appended → `self.changed = True`. The apply path
reads `rules_str` correctly, which is why the actual firewalld state is fine.

Both sides were introduced together in
0f8618657d414ae5d1ee30edbb2ec1b3909d358a (the in-memory backend for
`previous: replaced`), and it's in the current release.

## Minimal reproduction

```yaml
# repro.yml
- hosts: localhost
  become: true
  roles:
    - linux-system-roles.firewall
  vars:
    firewall:
      - zone: public
        rich_rule: 'rule family="ipv4" source address="192.0.2.0/24" reject'
        state: enabled
        permanent: true
```

```console
$ ansible-playbook repro.yml      # changed=1   (expected, first run)
$ ansible-playbook repro.yml      # changed=1   (BUG — expected changed=0)
$ sudo firewall-cmd --permanent --zone=public --list-rich-rules
rule family="ipv4" source address="192.0.2.0/24" reject   # identical across both runs
```

## Expected

The second run reports `changed: false`.

A fix probably just means aligning the key in `set_rich_rule()` (and the
`disabled` branch right below it) with `rules_str` — but I haven't traced
whether anything else keys off `rich_rules`, so I'll leave the exact patch to
you.

richm:
Cause: The in-memory backend was using a key 'rich_rule' for handling the rich
rules, but the code in get_config was using 'rules_str'.

Consequence: The firewall module added the rich rule every time to the 'rich_rule'
key, ignoring the 'rules_str' key, and was not idempotent.

Fix: Ensure that the key 'rich_rule' is used everywhere.  When reading the information
from the firewall api, ensure that the data under the key 'rules_str' is normalized
and added to the 'rich_rule' key.

Result: The firewall role can manage rich rules idempotently.

Fixes: https://github.com/linux-system-roles/firewall/issues/357

## Summary by CodeRabbit

* **Improvements**
  * Normalized firewall rich-rule handling so rules are stored and presented consistently as standardized strings across backends and configuration retrieval.

* **Tests**
  * Added/updated playbook checks to verify rich-rule presence and offline/online listing for firewall zones, including exact-match verification of an IPv4 reject rich rule.

### DIFF
--- a/library/firewall_lib.py
+++ b/library/firewall_lib.py
@@ -1948,6 +1948,8 @@ class InMemoryBackend:
                     zone_config["masquerade"] = masquerade
                     self.changed = True
 
+    # note: rich_rule is a single string, normalized using str(Rich_Rule(rule_str=original_string))
+    # the zone config field 'rich_rule' is a list of normalized strings
     def set_rich_rule(self, rich_rule):
         """Configure rich rules in a zone."""
         for config_type, zone_config in (
@@ -1957,12 +1959,12 @@ class InMemoryBackend:
             if config_type and zone_config is not None:
                 for item in rich_rule:
                     if self.state == "enabled":
-                        if item not in zone_config.get("rich_rules", []):
-                            zone_config.setdefault("rich_rules", []).append(item)
+                        if item not in zone_config.get("rich_rule", []):
+                            zone_config.setdefault("rich_rule", []).append(item)
                             self.changed = True
                     elif self.state == "disabled":
-                        if item in zone_config.get("rich_rules", []):
-                            zone_config["rich_rules"].remove(item)
+                        if item in zone_config.get("rich_rule", []):
+                            zone_config["rich_rule"].remove(item)
                             self.changed = True
 
     def set_source(self, source):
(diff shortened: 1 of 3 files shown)


richm:
[citest]

spetrosi (approved):

richm:
[citest]

PR #359, opened 2026-06-10
fix: ignore network device interface file that is not found

richm:
Cause: There seems to be a race condition in some cases where networkmanager
thinks there is a device but the system path is not found.

Consequence: The firewall role can issue an error when attempting to
manage a network interface by PCI ID.

Fix: If the file cannot be found, then it is not a valid device and
can be skipped, so skip it.

Result: The firewall role can manage network interfaces by PCI ID
without errors when there are race conditions.

Assisted-by: Cursor IDE using the models Composeer 2.5, Sonnet 4.6, Codex 5.3

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when collecting interface PCI data: unreadable or missing device info is skipped so overall interface collection continues.

* **Tests**
  * Added unit test verifying inaccessible PCI info is ignored and mappings include only successful reads.
  * Added an automated test debug step to display current PCI collection results during testing.

### DIFF
--- a/library/firewall_lib.py
+++ b/library/firewall_lib.py
@@ -2599,8 +2599,13 @@ def get_interface_pci():
         device_udi = nm_get_client().get_device_by_iface(interface).get_udi()
         device_path = os.path.join(device_udi, "device")
         for field in ["vendor", "device"]:
-            with open(os.path.join(device_path, field)) as _file:
-                interface_ids.append(_file.readline().strip(" \n")[2:])
+            try:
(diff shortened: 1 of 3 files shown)


richm:
[citest]

richm:
[citest]

spetrosi:
@coderabbitai review

richm:
[citest]

richm:
[citest]

PR #360, opened 2026-06-10
test: improve tests_ansible

richm:
There were some missing cleanup which caused problems with subsequent tests.

Use the role with previous replaced to ensure the machine is in the
correct state before testing, and for cleanup afterwards.

## Summary by CodeRabbit

* **Tests**
  * Refactored Ansible test playbook structure for improved validation of firewall module behavior.
  * Simplified test setup and cleanup logic, reducing redundant operations while maintaining comprehensive test coverage.


richm:
[citest]

PR #361, opened 2026-06-10
test: improve tests_reload_on_reset output

richm:
Use TEST_DEBUG=true by default so that we can see what happened if there is
a failure.

Name the output fields so we know what they are.

## Summary by CodeRabbit

**Tests**
- Updated test output format to explicitly emit key=value structured variables for improved parsing reliability and downstream integration.
- Enhanced test configuration with tag-based output parsing instead of fixed indices, delivering more deterministic behavior when debug mode is enabled.


richm:
[citest]

PR #365, opened 2026-06-26
fix: ensure firewalld is responding to requests after starting

richm:
Cause: The firewalld can be slow to start up on some platforms.

Consequence: The firewall module can be called before the server is responding and
will raise an error if it cannot connect to the firewall daemon.

Fix: Ensure that firewall-cmd --state returns "running" before calling
the module.

Result: The firewall role works even if the server is slow to startup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved firewall startup handling by waiting until the service is fully running before continuing, reducing the chance of follow-up tasks failing due to an unready firewall state.


richm:
[citest]

PR #384, opened 2026-08-19
feat: Add argument spec validation to Firewall role

DonatSzabo:
Enhancement: Added argument spec and assert role spec validation to the Firewall role. Also wrote tests for it found in tests/tests_invalid_input.

Reason: Because it is a good addition to the linux-system-roles project.

Result: Successfully added it and prepared tests for it. I used AI during this implementation.

Issue Tracker Tickets (Jira or BZ if any): https://github.com/linux-system-roles/postfix/issues/206 https://redhat.atlassian.net/browse/RHELMISC-16008

## Summary by CodeRabbit

* **New Features**
  * Added validation for firewall role settings, including supported keys, data types, and values.
  * Added clear error messages when invalid configuration is provided.
  * Documented available variables, defaults, accepted values, and example usage.

* **Tests**
  * Added coverage for invalid firewall configurations and module parameters to ensure they are rejected correctly.

### DIFF
--- a/meta/argument_specs.yml
+++ b/meta/argument_specs.yml
@@ -19,8 +19,10 @@ argument_specs:
       firewall:
         type: raw
         description: >
-          A list of firewall configuration dictionaries, or a single
-          dictionary. Each dictionary describes a desired firewall state
+          A list of firewall configuration dictionaries, a single
+          dictionary, or `null`. When set to `null` or an empty list,
+          the role gathers firewall facts without making changes. Each
+          dictionary describes a desired firewall state
           using one or more of the following keys:
           `service` - list of firewalld service name strings to add or
           remove from a zone. Service names must be defined in the
--- a/meta/argument_specs.yml
+++ b/meta/argument_specs.yml
@@ -91,6 +91,10 @@ argument_specs:
           existing settings. Defaults to `kept`;
           `includes` - list of service names to include in a custom
           service definition;
+          `detailed` - boolean to request detailed firewall facts. When
+          set to `true` with no other keys, the role gathers and returns
+          extended firewall configuration information in the
+          `firewall_config` fact;
           `firewalld_conf` - dictionary of firewalld.conf directives.
           Currently supports `allow_zone_drifting` (boolean) to set the
           AllowZoneDrifting directive.
--- a/meta/argument_specs.yml
+++ b/meta/argument_specs.yml
@@ -19,12 +19,12 @@ argument_specs:
     options:
       firewall:
         type: raw
+        default: []
         description: >
           A list of firewall configuration dictionaries, a single
           dictionary, or `null`.
           When set to `null` or an empty list, the role gathers firewall
           facts without making changes.
-          Defaults to an empty list.
           Each dictionary describes a desired firewall state
           using one or more of the following keys:
           `service` - list of firewalld service name strings to add or
@@ -105,14 +105,15 @@ argument_specs:
           AllowZoneDrifting directive.
       firewall_disable_conflicting_services:
         type: bool
+        default: false
         description: >
           Whether to disable and mask services that conflict with firewalld
-          such as iptables, nftables, and ufw. Defaults to `false`.
+          such as iptables, nftables, and ufw.
       firewall_transactional_update_reboot_ok:
         type: raw
+        default: null
         description: >
           Whether to permit reboots required by transactional updates. Set
           to `true` to allow reboots, `false` to disallow reboots, or
           leave unset so that the role will fail to ensure the reboot
           requirement is not overlooked. Accepts a boolean value or `null`.
-          Defaults to `null`.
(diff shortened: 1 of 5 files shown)


DonatSzabo:
[citest]

richm:
[citest_bad]

meta/argument_specs.yml
```diff
@@ -0,0 +1,116 @@
...
+        type: bool
+        description: >
+          Whether to disable and mask services that conflict with firewalld
+          such as iptables, nftables, and ufw. Defaults to `false`.
+      firewall_transactional_update_reboot_ok:
+        type: raw
+        description: >
+          Whether to permit reboots required by transactional updates. Set
+          to `true` to allow reboots, `false` to disallow reboots, or
+          leave unset so that the role will fail to ensure the reboot
```
spetrosi, line 118: @richm to me this sounds like this should be `false` by default, not null. Otherwise, what's the difference between null and false?

tests/tests_invalid_input.yml
```diff
@@ -0,0 +1,235 @@
...
+
+        - name: Run argument specs validation tests
+          when: ansible_version.full is version('2.10', '>=')
+          block:
+            - name: Argument specs reject invalid firewall_disable_conflicting_services
+              block:
+                - name: Run role with non-boolean firewall_disable_conflicting_services
+                  ansible.builtin.include_tasks: tasks/run_role_with_clear_facts.yml
+                  vars:
+                    firewall_disable_conflicting_services: not_a_bool
```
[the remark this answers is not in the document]
richm: `__sr_failed_when: false` is used when we want to ignore the role failure, typically during test cleanup.  In this case, we want the role to fail because we want to see if it detects the invalid input and reports an appropriate error, so we want the default behavior of `run_role_with_clear_facts.yml` which is `__sr_failed_when: true`

meta/argument_specs.yml
richm: https://github.com/linux-system-roles/firewall/blob/main/tasks/firewalld.yml#L52 needs to know if the user explicitly set it - so the value `null/none` means the user did not set the value.  If the value `is true` or `is false` this means the user explicitly set the value.

meta/argument_specs.yml
```diff
@@ -0,0 +1,116 @@
+# SPDX-License-Identifier: MIT
+---
+# https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#specification-format
+argument_specs:
+  main:
+    short_description: The firewall role.
+    description: >
+      The firewall role allows you to manage the firewall configuration
```
spetrosi, line 8: There is a concept of writing one sentence per line; this would make the diff much clearer for future edits. And the raw file is more readable in this format imo. Can you apply it here? AI should be able to just do this.

meta/argument_specs.yml
spetrosi: I see, if users set it to `false`, the role won't fail

richm:
please revert the changes to tests_service.yml - we will address this issue in a separate PR

richm:
[citest]

richm:
[citest]

richm:
rebase on top of `main` branch to pick up the fix in https://github.com/linux-system-roles/firewall/pull/386 for tests_service.yml

tests/tests_invalid_input.yml
```diff
@@ -0,0 +1,235 @@
...
+  hosts: all
+  tasks:
+    - name: Run invalid input tests
+      block:
+        # --- meta/argument_specs (invalid type) ---
+        # Role argument validation requires Ansible 2.10+; skipped on older
+        # versions.
+
+        - name: Run argument specs validation tests
+          when: ansible_version.full is version('2.10', '>=')
```
richm:
```suggestion
          when: ansible_version.full is version('2.11', '>=')
```

richm:
[citest]

richm:
[citest_bad]

tasks/assert_role_vars.yml
```diff
@@ -0,0 +1,81 @@
...
+    fail_msg: >-
+      firewall must be a list of dictionaries or a single dictionary,
+      got {{ firewall | type_debug }}
+  when: firewall is not none
+
+- name: Assert firewall dicts contain only valid keys
+  ansible.builtin.assert:
+    that:
+      - >-
+        item.keys() | list | difference(__firewall_valid_keys)
```
richm, line 20: use `idx` instead of `item` since you have `loop_index: idx`

DonatSzabo: I tried this but this leads to the role failing. The item variable in this case is the actual dict from the loop. Idx is used to identify the actual dict item number that is failing. Renaming item to idx would leat to idx being interpreted as a dict. How should I handle this? :)

richm: never mind - I was mistaking `loop_index` for `loop_var`

richm (dismissed):

richm:
[citest_all]

richm (approved):

PR #386, opened 2026-08-26
test: use __includes_and_helpers_support in tests_service.yml

richm:
If there is a task such as validate_argument_spec that needs access to `ansible_facts`
to evaluate parameter values, and this runs before the role can gather facts,
the facts will be undefined and you will get an error.  Ensure that this doesn't happen
by using a `set_fact` variable in the parameter values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for service configuration across supported Fedora-family systems.
  * Unsupported systems now correctly omit unavailable service includes and helper configuration.

### DIFF
--- a/tests/tests_service.yml
+++ b/tests/tests_service.yml
@@ -21,10 +21,7 @@
       destination:
         - 1.1.1.1
         - 1::1
-      includes: "{{ ['ssh', 'ldaps']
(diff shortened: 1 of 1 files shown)


richm:
[citest]

